PR #3, opened 2017-06-07
(MAINT) improve README description

kevpl:
[skip ci]

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,3 +6,50 @@ This includes all helper & installer methods.
 It might not be up to that state yet, but that's the goal for this library. If
 you see anything puppet-specific that you'd like to pull into this library out
 of beaker, please do, we would love any help that you'd like to provide. 
+
+# How Do I Use This?
+
+### Right Now? (beaker 3.x)
+
+At this point, beaker-puppet is included in beaker, so you don't have to _do_
+anything to get the methods in this library.
+
+You can use these methods in a test by referencing them by name without
+qualifications, as they're included in the beaker DSL by default.
+
+### In beaker's Next Major Version? (beaker 4.x)
+
+In beaker's next major version, the requirement for beaker-puppet will be pulled
+from that repo. When that happens, then the usage pattern will change. In order
+to use this then, you'll need to include beaker-puppet as a dependency right
+next to beaker itself.
+
+Once you've done that & installed the gems, in your test, you'll have to
+```ruby
+require 'beaker-puppet'
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,4 +5,51 @@ This includes all helper & installer methods.
...
+qualifications, as they're included in the beaker DSL by default.
+
+### In beaker's Next Major Version?
+
+In beaker's next major version, the requirement for beaker-puppet will be pulled
+from that repo. When that happens, then the usage pattern will change. In order
+to use this then, you'll need to include beaker-puppet as a dependency right
+next to beaker itself.
+
+Once you've done that & installed the games, in your test, you'll have to
```
tvpartytonight: Are we opinionated about if we want people to do this in their test files or in their helper somewhere? This might be a good place to endorse the usage of helpers to make the DSL available to the entirety of a test run, instead of individual tests.

Also, I think that you mean "installed the gems" not "installed the games". 👾  🎲  🎮

README.md
```diff
@@ -5,4 +5,51 @@ This includes all helper & installer methods.
...
 you see anything puppet-specific that you'd like to pull into this library out
-of beaker, please do, we would love any help that you'd like to provide. 
\ No newline at end of file
+of beaker, please do, we would love any help that you'd like to provide. 
+
+# How Do I Use This?
+
+### Right Now?
+
+At this point, beaker-puppet is included in beaker, so you don't have to _do_
```
tvpartytonight: Maybe we should specify the version 3.x here?

kevpl:
@tvpartytonight to your question about versions: great point, I've put version numbers in the section headers. Look good ❓ 

Fixed the typo (games -> gems), and to the larger question left there (from outdated folder):

    Are we opinionated about if we want people to do this in their test
    files or in their helper somewhere? This might be a good place to
    endorse the usage of helpers to make the DSL available to the
    entirety of a test run, instead of individual tests.

My opinion is that people should use the library code exactly where they need it. It will typically be from a test, but if they want to write their own helpers on top of the methods in the library, then they'll have to include it there too. I don't see the advantage of being opinionated about which place this goes in at this point. Am I just missing the advantage you're thinking of ❓

tvpartytonight:
Once a library is required in more than just a few files/tests, I think it makes sense to include at the helper level, since it reduces copy 🍝  from test to test. But the language here is fine, I just wanted to bring it up since I was thinking about it.

PR #6, opened 2017-06-28
(BKR-1118) add puppet5 install method

kevpl:
**Note** that this does not plug into `install_puppet` or `install_puppet_agent_on` methods at this point, it only provides basic puppet5 install functionality using the new `puppet5_install` method.

**Note** that that doesn't need to block merging this work, as those values need outside information that we don't have right now & can get in in a subsequent PR.

### DIFF
--- a/lib/beaker-puppet/install_utils/foss_utils.rb
+++ b/lib/beaker-puppet/install_utils/foss_utils.rb
@@ -972,9 +972,27 @@ module Beaker
           repo_config_folder_url = "%s/%s/%s/repo_configs/%s/" %
             [ buildserver_url, package_name, build_version, host.repo_type ]
 
-          repo = fetch_http_file( repo_config_folder_url,
-                                  repo_filename,
-                                  copy_dir )
+          repo_config_url = "#{ repo_config_folder_url }/#{ repo_filename }"
+          install_repo_configs_from_url( host, repo_config_url, copy_dir )
+        end
+
+        # Installs the repo configs on a given host
+        #
+        # @param [Beaker::Host] host Host to install configs on
+        # @param [String] repo_config_url URL to the repo configs
+        # @param [String] copy_dir Local directory to fetch files into & SCP out of
+        #
+        # @return nil
+        def install_repo_configs_from_url(host, repo_config_url, copy_dir = nil)
+          copy_dir ||= Dir.mktmpdir
+          repoconfig_filename = File.basename(  repo_config_url )
+          repoconfig_folder   = File.dirname(   repo_config_url )
+
+          repo = fetch_http_file(
+            repoconfig_folder,
+            repoconfig_filename,
+            copy_dir
+          )
 
           if host[:platform] =~ /cisco_nexus/
             to_path = "#{host.package_config_dir}/#{File.basename(repo)}"
--- a/lib/beaker-puppet/install_utils/puppet5.rb
+++ b/lib/beaker-puppet/install_utils/puppet5.rb
@@ -1,0 +1,171 @@
+require 'beaker/dsl/install_utils/foss_defaults'
+
+module BeakerPuppet
+  module Install
+    module Puppet5
+
+      include Beaker::DSL::InstallUtils::FOSSDefaults
+
+      # grab build json from the builds server
+      #
+      # @param [String] sha_yaml_url URL to the <SHA>.yaml file containing the
+      #    build details
+      #
+      # @return [Hash{String=>String}] build json parsed into a ruby hash
+      def fetch_build_details(sha_yaml_url)
+        dst_folder          = Dir.mktmpdir
+        sha_yaml_filename   = File.basename(  sha_yaml_url )
+        sha_yaml_folder_url = File.dirname(   sha_yaml_url )
+
+        sha_yaml_file_local_path = fetch_http_file(
+          sha_yaml_folder_url,
+          sha_yaml_filename,
+          dst_folder
+        )
+        file_hash = YAML.load_file( sha_yaml_file_local_path )
+
+        logger.debug("YAML HASH BELOW:")
+        logger.debug(file_hash)
+        logger.debug("PLATFORM_DATA BELOW:")
+        logger.debug(file_hash[:platform_data])
+        return sha_yaml_folder_url, file_hash[:platform_data]
+      end
+
+      # gets the artifact & repo_config URLs for this host in the build
+      #
+      # @param [Host] host Host to get artifact URL for
+      # @param [Hash] build_details Details of the build in a hash
+      # @param [String] build_url URL to the build
+      #
+      # @return [String, String] URL to the build artifact, URL to the repo_config
+      #   (nil if there is no repo_config for this platform for this build)
+      def host_urls(host, build_details, build_url)
+        packaging_platform = host[:packaging_platform]
+        if packaging_platform.nil?
+          message = <<-EOF
+            :packaging_platform not provided for host '#{host}', platform '#{host[:platform]}'
+            :packaging_platform should be the platform-specific key from this list:
+              #{ build_details.keys }
+          EOF
+          fail_test( message )
+        end
+
+        logger.debug("Platforms available for this build:")
+        logger.debug("#{ build_details.keys }")
+        logger.debug("PLATFORM SPECIFIC INFO for #{host} (packaging name '#{packaging_platform}'):")
+        packaging_data = build_details[packaging_platform]
+        logger.debug("- #{ packaging_data }, isnil? #{ packaging_data.nil? }")
+        if packaging_data.nil?
+          message = <<-EOF
+            :packaging_platform '#{packaging_platform}' for host '#{host}' not in build details
+            :packaging_platform should be the platform-specific key from this list:
+              #{ build_details.keys }
+          EOF
+          fail_test( message )
+        end
+
+        artifact_buildserver_path   = packaging_data[:artifact]
+        repoconfig_buildserver_path = packaging_data[:repo_config]
+        fail_test('no artifact_buildserver_path found') if artifact_buildserver_path.nil?
+
+        artifact_url    = "#{build_url}/#{artifact_buildserver_path}"
+        repoconfig_url  = "#{build_url}/#{repoconfig_buildserver_path}" unless repoconfig_buildserver_path.nil?
+        artifact_url_correct = link_exists?( artifact_url )
+        logger.debug("- artifact url: '#{artifact_url}'. Exists? #{artifact_url_correct}")
+        fail_test('artifact url built incorrectly') if !artifact_url_correct
+
+        return artifact_url, repoconfig_url
+      end
+
+      # install build artifact on the given host
+      #
+      # @param [Host] host Host to install artifact on
+      # @param [String] artifact_url URL of the project install artifact
+      # @param [String] project_name Name of project for artifact. Needed for OSX installs
+      #
+      # @return nil
+      def install_artifact_on(host, artifact_url, project_name)
+        variant, _, _, _ = host[:platform].to_array
+        onhost_package_file = nil
+        if variant == 'eos'
+          host.get_remote_file( artifact_url )
+          onhost_package_file = File.basename( artifact_url )
+        elsif variant == 'solaris'
+          artifact_filename = File.basename( artifact_url )
+          artifact_folder = File.dirname( artifact_url )
+          fetch_http_file( artifact_folder, artifact_filename, '.' )
+          onhost_package_dir = host.tmpdir( 'puppet_installer' )
+          scp_to host, artifact_filename, onhost_package_dir
+          onhost_package_file = "#{ onhost_package_dir }/#{ artifact_filename }"
+        elsif variant == 'osx'
+          on host, "curl -O #{ artifact_url }"
+          onhost_package_file = "#{ project_name }*"
+        end
+
+        if variant == 'eos'
+          # TODO Will be refactored into {Beaker::Host#install_local_package}
+          #   immediately following this work. The release timing makes it
+          #   necessary to have this here separately for a short while
+          host.install_from_file( onhost_package_file )
+        elsif onhost_package_file
+          if onhost_package_dir
+            host.install_local_package( onhost_package_file, '.' )
+          else
+            host.install_local_package( onhost_package_file )
+          end
+        else
+          host.install_package( artifact_url )
+        end
+      end
+
+      # Sets up the repo_configs on the host for this build
+      #
+      # @param [Host] host Host to install repo_configs on
+      # @param [String] repoconfig_url URL to the repo_config
+      #
+      # @return nil
+      def install_repo_configs_on(host, repoconfig_url)
+        if repoconfig_url.nil?
+          logger.warn("No repo_config for host '#{host}'. Skipping repo_config install")
+          return
+        end
+
+        install_repo_configs_from_url( host, repoconfig_url )
+      end
+
+      # Installs a specified puppet project on all hosts. Gets build information
+      #   from the provided YAML file located at the +sha_yaml_url+ parameter.
+      #
+      # @param [String] project_name Name of the project to install
+      # @param [String] sha_yaml_url URL to the <SHA>.yaml file containing the
+      #    build details
+      #
+      # @note This install method only works for Puppet versions >= 5.0
+      #
+      # @return nil
+      def install_from_build_data_url(project_name, sha_yaml_url)
+        if !link_exists?( sha_yaml_url )
+          message = <<-EOF
+            <SHA>.yaml URL '#{ sha_yaml_url }' does not exist.
+            Please update the `sha_yaml_url` parameter to the `puppet5_install` method.
+          EOF
+          fail_test( message )
+        end
+
+        base_url, build_details = fetch_build_details( sha_yaml_url )
+        hosts.each do |host|
+          artifact_url, repoconfig_url = host_urls( host, build_details, base_url )
+          if repoconfig_url.nil?
+            install_artifact_on( host, artifact_url, project_name )
+          else
+            install_repo_configs_on( host, repoconfig_url )
+            host.install_package( project_name )
+          end
+          configure_type_defaults_on( host )
+        end
+      end
+
+    end
+  end
+end
+
--- a/lib/beaker-puppet.rb
+++ b/lib/beaker-puppet.rb
@@ -13,6 +13,8 @@ end
   require "beaker-puppet/helpers/#{lib}_helpers"
 end
 
+require 'beaker-puppet/install_utils/puppet5'
+
 
 module BeakerPuppet
   module InstallUtils
@@ -23,6 +25,8 @@ module BeakerPuppet
     include Beaker::DSL::InstallUtils::FOSSUtils
     include Beaker::DSL::InstallUtils::EZBakeUtils
     include Beaker::DSL::InstallUtils::ModuleUtils
+
+    include BeakerPuppet::Install::Puppet5
   end
 
   module Helpers
--- a/lib/beaker-puppet/helpers/puppet_helpers.rb
+++ b/lib/beaker-puppet/helpers/puppet_helpers.rb
@@ -427,8 +427,9 @@ module Beaker
         #                      by the caller; this can be used for additional
         #                      validation, etc.
         #
-        # @return [Array<Result>, Result, nil] An array of results, a result object,
-        #   or nil. Check {#run_block_on} for more details on this.
+        # @return [Array<Result>, Result, nil] An array of results, a result
+        #   object, or nil. Check {Beaker::Shared::HostManager#run_block_on} for
+        #   more details on this.
         def apply_manifest_on(host, manifest, opts = {}, &block)
           block_on host, opts do | host |
             on_options = {}
--- a/spec/beaker-puppet/install_utils/puppet5_spec.rb
+++ b/spec/beaker-puppet/install_utils/puppet5_spec.rb
@@ -1,0 +1,282 @@
+require 'spec_helper'
+
+class ClassMixedWithDSLInstallUtils
+  include Beaker::DSL::Outcomes
+  include BeakerPuppet::Install::Puppet5
+
+  def logger
+    @logger ||= RSpec::Mocks::Double.new('logger').as_null_object
+  end
+end
+
+describe ClassMixedWithDSLInstallUtils do
+  let(:hosts)   { make_hosts( { :pe_ver => '3.0',
+                                      :platform => 'linux',
+                                      :roles => [ 'agent' ],
+                                      :type  => 'foss' }, 4 ) }
+
+  before :each do
+    allow( subject ).to receive( :hosts ) { hosts }
+  end
+
+
+  describe '#fetch_build_details' do
+    let( :platform_data ) { @platform_data || '' }
+    let( :deets ) { { :platform_data => platform_data } }
+
+    it 'sets & returns the relative folder' do
+      sha_yaml_folder = '/jams/of/the/rain'
+      sha_yaml_url    = "#{sha_yaml_folder}/puns.tgz"
+      expect( subject ).to receive( :fetch_http_file  )
+      expect( YAML    ).to receive( :load_file        ) { deets }
+
+      url, _ = subject.fetch_build_details( sha_yaml_url )
+      expect( url ).to be === sha_yaml_folder
+    end
+
+    it 'fetches & returns the build details' do
+      @platform_data = 'test of the man in the can carol?'
+      expect( subject ).to receive( :fetch_http_file  )
+      expect( YAML    ).to receive( :load_file        ) { deets }
+
+      _, hash = subject.fetch_build_details( 'sha_yaml_url' )
+      expect( hash ).to be === @platform_data
+    end
+
+    it 'stores the file in a good location' do
+      allow( YAML ).to receive( :load_file ) { deets }
+
+      correct_location = '/my/fake/test/dir'
+      expect( Dir ).to receive( :mktmpdir ) { correct_location }
+      expect( subject ).to receive( :fetch_http_file ).with(
+        anything, anything, correct_location
+      )
+      subject.fetch_build_details( 'sha_yaml_url' )
+    end
+
+  end
+
+  describe '#host_urls' do
+
+    let( :artifact_path       ) { @artifact_path      || ''             }
+    let( :repo_config         ) { @repo_config        || nil            }
+    let( :packaging_platform  ) { @packaging_platform || 'el-7-x86_64'  }
+    let( :host                ) {
+      host = hosts[0]
+      allow( host ).to receive( :[] ).with( :packaging_platform ) {
+        packaging_platform
+      }
+      host
+    }
+    let( :build_details   ) {
+      details = {
+        packaging_platform => {
+          :artifact     => artifact_path,
+          :repo_config  => repo_config
+        }
+      }
+      details
+    }
+
+    it 'fails if there\'s no artifact value for the given platform' do
+      allow( artifact_path ).to receive( :nil? ) { true }
+      expect {
+        subject.host_urls( host, build_details, '' )
+      }.to raise_error(
+        Beaker::DSL::Outcomes::FailTest,
+        /^no artifact.*path found$/
+      )
+    end
+
+    it 'fails if the artifact_url doesn\'t exist' do
+      allow( subject ).to receive( :link_exists? ) { false }
+      expect {
+        subject.host_urls( host, build_details, '' )
+      }.to raise_error(
+        Beaker::DSL::Outcomes::FailTest,
+        /^artifact url.*incorrectly$/
+      )
+    end
+
+    it 'fails if the host doesn\'t have a packaging_platform' do
+      allow( packaging_platform ).to receive( :nil? ) { true }
+      allow( host ).to receive( :[] ).with( :platform ) { 'fake-platform' }
+      expect {
+        subject.host_urls( host, build_details, '' )
+      }.to raise_error(
+        Beaker::DSL::Outcomes::FailTest,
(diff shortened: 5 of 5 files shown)


lib/beaker-puppet/install_utils/puppet5.rb
```diff
@@ -0,0 +1,109 @@
...
+        end
+      end
+
+      # install puppet5 project on all hosts
+      #
+      # @param [String] project_name Name of the project to install
+      # @param [String] project_sha SHA ref of the project to install
+      #
+      # @return nil
+      def puppet5_install(project_name, project_sha)
```
glennsarti: Does this mean we need to change our helper invocations to `puppet5_install` in order to use puppet 5 instead of puppet 4?

glennsarti: Oh nvm...read the PR description

lib/beaker-puppet/install_utils/puppet5.rb
```diff
@@ -0,0 +1,109 @@
...
+      # @return [String] URL to the build artifact
+      def hosts_artifact_url(host, build_details, build_url)
+        platform_key = host[:platform]
+        variant, version, arch, _ = platform_key.to_array
+        # step: map platform strings into platform_key's for the hash
+        if variant == 'debian' && arch == 'x86_64'
+          platform_key.sub!('x86_64', 'amd64')
+        elsif variant == 'fedora'
+          platform_key.sub!(version, "f#{version}")
+        end
```
joshcooper: Can we do this translation in packaging instead of beaker, so that beaker stays dumb?

lib/beaker-puppet/install_utils/puppet5.rb
```diff
@@ -0,0 +1,109 @@
...
+          artifact_folder = File.dirname( artifact_url )
+          fetch_http_file( artifact_folder, artifact_filename, '.' )
+          scp_to host, artifact_filename, '/tmp'
+          onhost_package_file = "/tmp/#{ artifact_filename }"
+        elsif variant == 'debian'
+          onhost_package_file = '/tmp/puppet.deb'
+          on host, "wget -O #{ onhost_package_file } #{ artifact_url }"
+        elsif variant == 'osx'
+          on host, "curl -O #{ artifact_url }"
+          onhost_package_file = "#{ project_name }*"
```
joshcooper: Do beaker's `Host` classes already have a method for downloading/fetching a file, e.g. `host.download_from_url(artifact_url)`, so that we don't need if/then/else, and knowledge about wget vs curl?

lib/beaker-puppet/install_utils/puppet5.rb
kevpl: @joshcooper I don't see why not, but when we originally got together with @mwaggett, I told her to just use her normal packaging platform keys, since it's arbitrary which side of the packaging-beaker conversation figures them out.

That being said, this info is all OS-specific variation, which makes me think that the packaging repo would be a better place to handle it. Now that we've figured it out though, it seems like we could take our time moving that over.

@mwaggett does this make sense?

lib/beaker-puppet/install_utils/puppet5.rb
kevpl: hm, I'm not seeing a method that abstracts between these two right off the bat.

There's a `fetch_http_file` method that we often use with `scp_to` to install PE artifacts ([`install_repo_configs` example](https://github.com/puppetlabs/beaker-puppet/blob/master/lib/beaker-puppet/install_utils/foss_utils.rb#L966-L975)), but if we can get them right to the SUT rather than having to go through the coordinator that would be more efficient.

This pattern is pretty typical in the FOSS installs (just straight using `on` to execute curl or wget), but we don't typically see it side-by-side because of the way the work is split up in the install code (historically, we've done _everything_ for a platform in one body of a switch statement, rather than break up the tasks of download artifact vs install artifact). Refactoring this work into a helper method seems like a useful thing, but something that can be done separately from this work.

@joshcooper am I just missing the method, or do you think creating that helper method should block this PR?

lib/beaker-puppet/install_utils/puppet5.rb
mwaggett: @kevpl It kind of doesn't make sense to me for packaging to need to know how your test hosts are named, but I am open to discussing this. That being said, we do have a [ticket](https://tickets.puppetlabs.com/browse/RE-8699) for unifying platform naming (though we probably won't get to that for a while). 

I think the issue is that whenever one of us changes a platform tag (e.g. we're about to drop the 'f' from fedora versions), it will require a change in the other place. Another thing that will be solved by the metadata service, when we get to it..

kevpl: @mwaggett these aren't dealing with details of how test hosts are named, but with inconsistencies in the keys provided in the `platform_data` data structure.

Glad to hear that fedora's f-version prefix will be going away soon. Packaging provides two different 64-bit architecture strings: `x86_64` or `amd64`. I assume that this distinction means something for packaging? For beaker, it doesn't care between them & so we've consistently named them `x86_64`. Even when we move to the metadata service, the problem will exist until we agree on what a consistent naming standard.

mwaggett: @kevpl We use `amd64` for debuntus, I think to be consistent with how they do it upstream.

How strongly do you and @joshcooper feel about moving that logic to packaging? I'm happy to have a conversation about it, but I kind of don't see a huge benefit.

kevpl: @mwaggett I'm not sure how @joshcooper feels about it, but as I'm seeing it, the question is really about the tradeoffs between either:

- being consistent across the OSes
- being consistent internally to Puppet's tooling

If we're consistent across the OS vendors, then we have to make sure that OS-specific changes are propagated across all of our tooling (which means across teams). Then, when those details change (& we can't control when that might happen), we have to update all of our projects to fit.

If we're consistent across internal Puppet tooling, then packaging becomes the place where the mapping between OS-specific details & puppet tooling takes place. Packaging would be the [adapter](https://en.wikipedia.org/wiki/Adapter_pattern) between Puppet (the company) & OS vendors. Then if something changes, the adapter can be updated by our OS-specific experts internally, & no one else has to miss a beat.

Note that we're off on a tangent of future work here. This doesn't need to block this PR merging, as pulling this case statement out would be trivial to do in the future, the only concern being a little scheduling between beaker-puppet & packaging work changes.

lib/beaker-puppet/install_utils/puppet5.rb
```diff
@@ -0,0 +1,109 @@
...
+
+      # grab build json from the builds server
+      #
+      # @param [String] project_name Name of the project
+      # @param [String] project_sha SHA ref of the build for the project
+      #
+      # @return [String, String] URL to the build for the SHA of the project,
+      #   build json parsed into a ruby hash
+      def fetch_build_details(project_name, project_sha)
+        buildserver_url = 'http://builds.delivery.puppetlabs.net'
```
tvpartytonight: Does this need to be configurable for users outside of the puppet firewall?

lib/beaker-puppet/install_utils/puppet5.rb
mwaggett: yeah, I agree that internal consistency with packaging as an interface seems most desirable, but that is future work. Mostly I'm just eager to get this stuff in now. 😄

lib/beaker-puppet/install_utils/puppet5.rb
```diff
@@ -0,0 +1,109 @@
...
+        buildserver_projectsha_url = "#{buildserver_url}/#{project_name}/#{project_sha}"
+        file_folder = "#{buildserver_projectsha_url}/artifacts"
+        filename = "#{project_sha}.yaml"
+        dst_folder = Dir.pwd
```
tvpartytonight: Is there a reason we put this in the local directory, as opposed to using `Dir.mktmpdir`?

lib/beaker-puppet/install_utils/puppet5.rb
```diff
@@ -0,0 +1,109 @@
...
+
+        # fetch & parse file
+        file_local_path = fetch_http_file(file_folder, filename, dst_folder)
+        file_hash = YAML.load_file(file_local_path)
+
+        # getting the relevant data from the file
+        logger.perf_output("YAML HASH BELOW:")
```
tvpartytonight: Why are these `perf_output` instead of `debug` or `info`?

lib/beaker-puppet/install_utils/puppet5.rb
joshcooper: Ideally, we shouldn't open-code the logic for what it means to execute curl/wget across all platforms, but that's always been an issue in beaker. So I don't think it should block this PR.

lib/beaker-puppet/install_utils/puppet5.rb
joshcooper: 👍  to use `mktmpdir`, 1) because it's a secure directory, and 2) because it will be unique for each run, so any build artifacts from a previous run, e.g. `puppet-agent-latest-x64.msi`, won't pollute future runs.

lib/beaker-puppet/install_utils/puppet5.rb
joshcooper: Sorry to be the downer at the party, but don't we need to fix the naming issue before making these changes? For example, take the fedora case. In beaker+packaging v1, we have an acceptance job with `TEST_TARGET=fedora25-64a`. It's run through `beaker-hostgenerator` to generate a host config like:

```yaml
---
HOSTS:
  fedora25-64-1:
    hypervisor: vmpooler
    platform: fedora-25-x86_64
    template: fedora-25-x86_64
    roles:
    - agent
```

Beaker uses `platform=fedora-25-x86_64` internally. With this PR, we magically change it to `fedora-f25-x86_64`. We then use that key to lookup the build artifact that the packaging repo created.

In the future, packaging v2 will no longer include `f` in the key. Now we need to update beaker to match. But beaker needs to be able to install packages created by v1 and v2 of packaging, e.g. in the puppet-agent LTS pipeline (1.10.x) and the latest 5.x pipeline. Since the beaker change will break older pipelines, we have to take a semver major, or add a feature flag that consumers can opt-into, or add logic to try `f25` and fallback to `25`. In other words, we find ourselves reimplementing the same set of hacks in beaker, that we're trying to avoid in the beaker+packaging epic.

In my mind, beaker should be data-driven. So given a `key`, lookup the packaging artifact/repo in the hash produced by the packaging repo. So long as the key doesn't change for a given OS/arch combo, then we can make changes in the packaging layout without breaking any consumer, e.g. `fc25`. But because this is a bootstrapping problem, we have to get the name right to begin with. I don't think the key necessarily needs to be `platform`, eg it could be something like `host[:packaging_name]`, etc

mwaggett: @joshcooper @kevpl so if I'm understanding correctly, you would like the fedora platform tags to always be of the form fedora-<ver>-<arch> and to always use x86_64 instead of a mix of that and amd64? The former should be easy; I will look into what it would take for the latter.

kevpl:
@tvpartytonight thanks for the 👀 , I think most of those were just me not cleaning that particular method up for the PR, so thanks for the catch!

@joshcooper I had a wrong idea of how versions would work across both repos, so thanks for bringing up that situation! I like the data-driven bits & how we can get around dealing with the changes from the beaker side using a different property than platform-munging, so I've gone that route in the latest update! Let me know if there are any parts of that objection that still stand, or if you have any feedback on the new changes.

@mwaggett, this should make beaker work with changes across the packaging repo, as we've moved dealing with that to the host's configuration rather than dealing with it in beaker code. I'm glad we had the conversation about naming since it'll be important when we move to reconcile these though, so thanks for your help!

kevpl:
@joshcooper @mwaggett I've updated this PR with the changes we talked about from Friday: making `puppet5_install` take a full URL rather than parameters & building the URL out of it.

Note that my local testing fails, but that's because of the relative to the <SHA>.yaml file work that hasn't come through to the builds server that I'm seeing. I'll give an example of this in the BKR ticket, so we can be confident that this behaves correctly.

mwaggett:
@kevpl These packages were built with the changes we discussed: http://builds.delivery.puppetlabs.net/puppet-agent/ff3e02ddf16781f94b50918b995c6e793597f9a9/
(https://github.com/puppetlabs/packaging/pull/690)

kevpl:
@mwaggett thank for that ref! I'm able to test it & see that it goes green with my newest code from yesterday.

Forgot to change the name from `packaging_name` to `packaging_platform` yesterday, so I did that this morning. @mwaggett @joshcooper this is ready for review

mwaggett:
since I'm not familiar with this codebase, I don't know how valuable a review from me is, but if you're able to fetch and test the builds I made with my PR, then this looks good to me!

lib/beaker-puppet/install_utils/puppet5.rb
```diff
@@ -0,0 +1,128 @@
...
+          on host, "wget -O #{ onhost_package_file } #{ artifact_url }"
+        elsif variant == 'osx'
+          on host, "curl -O #{ artifact_url }"
+          onhost_package_file = "#{ project_name }*"
+        end
+
+        if onhost_package_file
+          host.install_local_package( onhost_package_file )
+        else
+          host.install_package( artifact_url )
```
joshcooper: Trying this on ubuntu-16.04-amd64, this method ends up calling:

```
q2p1dby68kwqdaq.delivery.puppetlabs.net (q2p1dby68kwqdaq.delivery.puppetlabs.net) 14:06:34$ apt-get install --force-yes  -y http://builds.delivery.puppetlabs.net/puppet-agent/f0005a296b2e1f5d6a7541aac005d41550690bc5/artifacts/./deb/xenial/PC1/puppet-agent_5.0.0.137.gf0005a2-1xenial_amd64.deb
    Reading package lists...
    Building dependency tree...
    Reading state information...
    E: Unable to locate package http://builds.delivery.puppetlabs.net/puppet-agent/f0005a296b2e1f5d6a7541aac005d41550690bc5/artifacts/./deb/xenial/PC1
    E: Couldn't find any package by glob 'http://builds.delivery.puppetlabs.net/puppet-agent/f0005a296b2e1f5d6a7541aac005d41550690bc5/artifacts/./deb/xenial/PC1'
    E: Couldn't find any package by regex 'http://builds.delivery.puppetlabs.net/puppet-agent/f0005a296b2e1f5d6a7541aac005d41550690bc5/artifacts/./deb/xenial/PC1'
```

But it fails because `apt-get` can't seem to install the `.deb` from a URL. Should we curl/wget the `deb` locally to the SUT first?

lib/beaker-puppet/install_utils/puppet5.rb
```diff
@@ -0,0 +1,128 @@
...
+      end
+
+      # install puppet5 project on all hosts
+      #
+      # @param [String] sha_yaml_url URL to the <SHA>.yaml file containing the
+      #    build details
+      # @param [String] project_name Name of the project to install
+      #
+      # @return nil
+      def puppet5_install(project_name, sha_yaml_url)
```
joshcooper: If this method should be usable for projects like `puppetserver`/`puppetdb` that have dependencies, then I think we always want to install the `repo_config` for platforms with package managers, and only install the package directly for things like msi/dmg/etc.

Also I'm thinking this isn't puppet5 specific, more like `install_from_build_data_url`?

lib/beaker-puppet/install_utils/puppet5.rb
kevpl: @joshcooper added ubuntu to the local install case. Which job should I be looking to to complete the platform array for this install method?

kevpl:
@joshcooper I've added `repo_config` installation as a step.

I've changed the `puppet_install` name as well & added a comment to note that it was made in puppet 5 for deprecation purposes later.

lib/beaker-puppet/install_utils/puppet5.rb
joshcooper: These are the current list of BHG platforms we test against: https://github.com/puppetlabs/ci-job-configs/blob/626f1617cca7814c0b01299998573aa949e3f67e/jenkii/jenkins-master-prod-1/projects/puppet-agent.yaml#L207-L231

lib/beaker-puppet/install_utils/puppet5.rb
```diff
@@ -0,0 +1,162 @@
...
+        variant, _, _, _ = host[:platform].to_array
+        onhost_package_file = nil
+        if variant == 'solaris'
+          artifact_filename = File.basename( artifact_url )
+          artifact_folder = File.dirname( artifact_url )
+          fetch_http_file( artifact_folder, artifact_filename, '.' )
+          scp_to host, artifact_filename, '/tmp'
+          onhost_package_file = "/tmp/#{ artifact_filename }"
+        elsif variant =~ /^(debian|ubuntu)$/
+          onhost_package_file = '/tmp/puppet.deb'
```
joshcooper: Can we use `host.tmpfile('puppet.deb')` to ensure we don't have conflicts?

melissa:
Hey all! Any movement on this?

kevpl:
Hey @melissa, I'm working on getting the platform coverage that we're looking for, but I'm hitting some difficulties testing the different platforms with the vmpooler degredation we're experiencing, so that's slowing this effort down. I'll be keeping an eye on that, & coming back to it as soon as I can (hopefully in the next day or so).

melissa:
Thank you so much for the update, @kevpl ! Let me know if I can be of any help

lib/beaker-puppet/install_utils/puppet5.rb
kevpl: @joshcooper working on testing these platforms, @melissa noticed that ubuntu1610 was on the list & thought we weren't shipping puppet-agent on that platform right now. Checking with CI peeps, it sounds like the up-to-date version of the platform array would just be the file on the [master branch](https://github.com/puppetlabs/ci-job-configs/blob/master/jenkii/jenkins-master-prod-1/projects/puppet-agent.yaml#L157-L180) for those pipelines, but it looks like that has ubuntu1610 as well. What would be the right version of that file that we should be testing from to get the correct array of platforms?

mwaggett: @kevpl ubuntu1610 has been EOL'd for puppet5 and beyond, but is currently still being shipped in the 1.10.x series. I think the plan is to EOL it there too, but that hasn't been done yet.

mwaggett: https://github.com/puppetlabs/build-data/blob/release/builder_data.yaml#L64 is a pretty solid list (and I think currently accurate), but sometimes we forget to remove EOL'd things.

kevpl: thanks for clearing this up @mwaggett! 🥇

melissa: Hey all! We're trying to have a single source of truth for the supported platforms. Anything in github should be based on the confluence page, but should not be considered a final source of truth. For the final source of truth, consult https://confluence.puppetlabs.com/display/PROD/Currently+Supported+OS+Platforms. However, it does look like it has out of date information for yakkety :(

A lot of this decision not to ship Yakkey comes from the following email conversation between Ruth, @joshcooper and @haus 

>>> (Note: probably only the platform team cares about this, but I don't know what the appropriate email list is for that that definitely includes the people who care about this and I already spent 5 minutes trying to figure it out before giving up.)

>>>The plaftorm team has done some work to get puppet-agent and puppet server packages for Yakkety (Ubuntu 16.10) for Puppet 5... but there are still bugs to sort out causing redness in pipelines.

>>Curious if the bugs are specific to Yakkety-only, or will we hit the same issues as soon as we try to add support for Zesty 17.04 in https://tickets.puppetlabs.com/browse/PA-873?

>Yakkety has the weird puppet-agent package that isn't from us, while Zesty does not, fwiw.
 
>>>Given that the current date for GA for Puppet 5 is end of June and Yakkety goes EOL in July, it seems like rather than trying to spend more time getting this to work, our time could be better spent elsewhere.

>>>I talked with Eric0 briefly and he was on board with not shipping Yakkety packages for Puppet platform 5, but I wasn't sure what further things (confluence pages??) need updating, and also wanted to make sure this discussion was widely circulated.

kevpl:
@joshcooper @melissa updated w/`tmpfile` improvements & additional platform testing. `eos` was the only platform that needed additional logic, & some of this will be improved after this work is released. Apologies with the delay, there was a snag dealing with some `cisco` testing, but now we're through that. Let me know if you have any questions or comments about this.

lib/beaker-puppet/install_utils/puppet5.rb
```diff
@@ -0,0 +1,176 @@
...
+            Please update the `sha_yaml_url` parameter to the `puppet5_install` method.
+          EOF
+          fail_test( message )
+        end
+
+        base_url, build_details = fetch_build_details( sha_yaml_url )
+        hosts.each do |host|
+          artifact_url, repoconfig_url = host_urls( host, build_details, base_url )
+          install_repo_configs_on( host, repoconfig_url )
+          install_artifact_on( host, artifact_url, project_name )
```
melissa: This kind of looks like we're installing the repo_configs if they're available, but then not using those enabled repos? In `install_artifact_on`, it looks like we're just installing the artifact directly from the `artifact` url. Am I misinterpreting this?

kevpl:
@melissa I've pushed new changes that take advantage of the repo_config info if it exists. Let me know if you see any other issues with this. Thank you!

lib/beaker-puppet/install_utils/puppet5.rb
```diff
@@ -0,0 +1,177 @@
...
+          #   necessary to have this here separately for a short while
+          host.install_from_file( onhost_package_file )
+        elsif onhost_package_file
+          if onhost_package_dir
+            host.install_local_package( onhost_package_file, '.' )
+          else
+            host.install_local_package( onhost_package_file )
+          end
+        else
+          host.install_package( artifact_url )
```
melissa: So this will give the whole artifact url as defined in the yaml file, yes? Like `http://builds.puppetlabs.lan/puppet-agent/1.9.1/artifacts/el/7/PC1/x86_64/puppet-agent-1.9.1-1.el7.x86_64.rpm` or `http://builds.puppetlabs.lan/puppet-agent/903bd6cbb8bfc6f88a6be61acc86b8bfa12b8690/artifacts/deb/xenial/PC1/puppet-agent_5.0.1.113.g903bd6c-1xenial_amd64.deb`?

I'm not 100% sure all package management sysmtes are going to be able to handle that as an argument, unless it's getting parsed somewhere that I've missed

```
root@plb4iyu3cgus7pk:~# apt-get install --force-yes -y http://builds.puppetlabs.lan/puppet-agent/903bd6cbb8bfc6f88a6be61acc86b8bfa12b8690/artifacts/deb/xenial/PC1/puppet-agent_5.0.1.113.g903bd6c-1xenial_amd64.deb
Reading package lists... Done
Building dependency tree
Reading state information... Done
E: Unable to locate package http://builds.puppetlabs.lan/puppet-agent/903bd6cbb8bfc6f88a6be61acc86b8bfa12b8690/artifacts/deb/xenial/PC1
E: Couldn't find any package by glob 'http://builds.puppetlabs.lan/puppet-agent/903bd6cbb8bfc6f88a6be61acc86b8bfa12b8690/artifacts/deb/xenial/PC1'
E: Couldn't find any package by regex 'http://builds.puppetlabs.lan/puppet-agent/903bd6cbb8bfc6f88a6be61acc86b8bfa12b8690/artifacts/deb/xenial/PC1'
```

lib/beaker-puppet/install_utils/puppet5.rb
```diff
@@ -0,0 +1,177 @@
...
+        end
+
+        base_url, build_details = fetch_build_details( sha_yaml_url )
+        hosts.each do |host|
+          artifact_url, repoconfig_url = host_urls( host, build_details, base_url )
+          if repoconfig_url.nil?
+            install_artifact_on( host, artifact_url, project_name )
+          else
+            install_repo_configs_on( host, repoconfig_url )
+            host.install_package( project_name )
```
melissa: 💯

lib/beaker-puppet/install_utils/puppet5.rb
melissa: Oh, wait, this might not matter, right? So, this method will only get called if repo_configs is nil, so things without a repo (osx, eos, etc) are the only platforms that should be using this method unless for some reason things broke and those repo_configs weren't created...

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -963,9 +963,27 @@ def install_repo_configs(host, buildserver_url, package_name, build_version, cop
...
+        # @param [String] repo_config_url URL to the repo configs
+        # @param [String] copy_dir Local directory to fetch files into & SCP out of
+        #
+        # @return nil
+        def install_repo_configs_from_url(host, repo_config_url, copy_dir = nil)
+          copy_dir ||= Dir.mktmpdir
+          repoconfig_filename = File.basename(  repo_config_url )
+          repoconfig_folder   = File.dirname(   repo_config_url )
+
+          repo = fetch_http_file(
```
melissa: Is there a graceful way to fail back to installing the package directly if for some reason something broke and the repo_config files aren't available? That actually might not be something we want to allow beaker to do, now that I'm thinking about it....

lib/beaker-puppet/install_utils/puppet5.rb
kevpl: @melissa yes, this code path will only get called with full artifact URLs when repo_configs aren't set.

In terms of your error question, one of the goals of these changes is that we want beaker to remain dumb & not able to know when repo_configs should & shouldn't be available, so testing for this would be difficult unless we broke that. It seems like the packaging system that generates the repo_configs would be better suited to validating repo_config existence before saying a build is finished.

I can see how validating & behaving nicely on both sides would be a nice experience for the end user, but I think that's how we often end up with poor duplicate copies between beaker & the packaging system. 

Right now the [output of a failure case](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/helpers/web_helpers.rb#L63-L67) is pretty explicit about what caused it to fail if these are unreachable, so I don't think it's hard to debug. There are some other options for dealing with failure cases, but I think we'll talk about them in the other comment that's on the specific lines in question.

melissa:
I think this looks great, and I can't wait to see it in action!

lib/beaker-puppet/install_utils/foss_utils.rb
kevpl: @melissa I'm wary of degrading to a different install if the repo_configs aren't available. That's something that we've done historically, and we've had two problems with that approach: transparency & debugging/maintenance.

One is that magic changes underneath a user (like an auto install-method switch) is much harder to debug & understand. The only way to be transparent about degrading in this fashion that we've seen is to scream into the console log, but then we face the problem of screaming wolf too many times & nobody pays attention to the log colors/warnings/errors anymore.

Another is that the more avenues we use to install, the harder it is to maintain that code. If the issue is potentially in the packaging system such that the symptom is that repo_config files aren't being made available, then finding alternate ways to install creates more distance between the cause of errors & the way that we find them. It seems that failing at build-time would be even better, but failing as early as possible during the install step is the best that beaker can do to help a user find a bug like that. 

If we're thinking of transient issues, like network blips over getting the appropriate file rather than the file being in the right place, then there are some things we can definitely do to help with that. Our usual suggestion at this point would be to back off & retry, hoping that those network issues resolve themselves over a small break. Unfortunately, this can hide issues & delay job completion, but it can help overcome things like this. I would say at this point that there's no reason to add that complexity if we don't need it though.

melissa: I agree those are good reasons to be wary of that course of action. I think we should move forward with this as is

kevpl:
@joshcooper @tvpartytonight if you have some time, I'd appreciate another look at this.

melissa:
@joshcooper @tvpartytonight I would love to get this in soon so we can start leveraging it for those projects targeted at puppet5!

melissa:
Also, we have a good opportunity to update the beaker version in the puppet agent 5.x branch tomorrow after the latest release. I'd love to get the tests going and suss out any potential issues we might run into

kevpl:
@puppetlabs/beaker it would be great to get some more eyes on this too!

melissa:
Hey @puppetlabs/beaker @joshcooper @tvpartytonight puppet agent 5.1.0 went out yesterday, so we have a great window to bump the beaker version in that branch!

melissa:
@kevpl it looks like there's a merge conflict that needs to be resolved :(

kevpl:
merge conflict resolved

kevpl:
The mirror didn't have oga from Rishi's PR today, so I pushed it up.

@puppetlabs-jenkins retest this please.

kevpl:
same with ruby-ll, ugh.

tvpartytonight:
whoooooo!

melissa:
YAY!

lib/beaker-puppet/install_utils/foss_utils.rb
melissa: A little followup, yes, the error message here when this file is missing is very informative. But here's a question I should have asked earlier: why not derive the repo_config info from the yaml file? Right now we're guessing what the repo name should be, which could be wrong
```
           repo_filename = host.repo_filename( package_name, build_version )
            repo_config_folder_url = "%s/%s/%s/repo_configs/%s/" %
              [ buildserver_url, package_name, build_version, host.repo_type ]
```

melissa: hmm, nope, this is totally as it should be. The new functionality should be and is isolated. My bad. IGNORE ME.

PR #7, opened 2017-06-28
(BKR-1147) Fail to pkg install on ec2 for dev-repo

johnduarte:
This commit modifies the `install_puppet_agent_dev_repo_on` method
to install the puppet-agent package if the host is on ec2 and el.
This is done because the ec2 instance does not have access to the
default dev repo.


kevpl:
@johnduarte looks OK to me.

kevpl:
@johnduarte would be releasing right now, but it looks like automation is in maintenance mode.

Will release as soon as we can.

PR #8, opened 2017-07-10
(MAINT) remove `dev_builds_repos` mentions

kevpl:
@melissa noticed while updating downstream
projects that `dev_builds_repos` was an option
beaker declared as necessary, but wasn't in use.
it looks like it was removed in BKR-648
(github.com/puppetlabs/beaker/pull/1044), but
these mentions lingered. This PR removes them,
so hopefully we won't get as confused next time.


tvpartytonight:
Not seen in beaker-pe or beaker as well; 👍 !

PR #9, opened 2017-07-10
(maint) Update beaker to use newest puppet repos

melissa:
This commit updates beaker to use the new puppet5 repos. However, the
repos that this commit updates beaker to use are the rolling repos.
These are links to the currently active major version series. When
puppet6 is out and stable, we will switch these puppet repos to link to
the puppet6 repos. We will update these links for every new major
version series. This means that we will not have to make any changes in
beaker unless we want to pin a project to a particular version series,
i.e., 5. Ideally we wouldn't need to pin, but that's not super realistic

### DIFF
--- a/lib/beaker-puppet/install_utils/foss_utils.rb
+++ b/lib/beaker-puppet/install_utils/foss_utils.rb
@@ -331,7 +331,7 @@ module Beaker
         # @option opts [String] :puppet_gem_version Version of puppet to install via gem if no puppet-agent package is available
         # @option opts [String] :mac_download_url Url to download msi pattern of %url%/puppet-agent-%version%.msi
         # @option opts [String] :win_download_url Url to download dmg pattern of %url%/puppet-agent-%version%.msi
-        # @option opts [String] :puppet_collection Defaults to 'pc1'
+        # @option opts [String] :puppet_collection Defaults to nil
         # @option opts [Boolean] :run_in_parallel Whether to run on each host in parallel.
         #
         # @return nil
@@ -339,7 +339,6 @@ module Beaker
         # @raise [FailTest] When error occurs during the actual installation process
         def install_puppet_agent_on(hosts, opts = {})
           opts = FOSS_DEFAULT_DOWNLOAD_URLS.merge(opts)
-          opts[:puppet_collection] ||= 'pc1' #hi!  i'm case sensitive!  be careful!
           opts[:puppet_agent_version] ||= opts[:version] #backwards compatability with old parameter name
 
           run_in_parallel = run_in_parallel? opts, @options, 'install'
@@ -438,11 +437,7 @@ module Beaker
         # @api private
         def install_puppet_from_rpm_on( hosts, opts )
           block_on hosts do |host|
-            if host[:type] == 'aio'
-              install_puppetlabs_release_repo(host,'pc1',opts)
-            else
-              install_puppetlabs_release_repo(host,nil,opts)
-            end
+            install_puppetlabs_release_repo(host,opts[:puppet_collection],opts)
 
             if opts[:facter_version]
               host.install_package("facter-#{opts[:facter_version]}")
@@ -703,12 +698,15 @@ module Beaker
         # @param [Hash{Symbol=>String}] opts An options hash
         # @option opts [String] :puppet_agent_version The version of Puppet Agent to install, defaults to latest
         # @option opts [String] :mac_download_url Url to download msi pattern of %url%/puppet-%version%.dmg
-        # @option opts [String] :puppet_collection Defaults to 'PC1'
+        # @option opts [String] :puppet_collection Defaults to nil
         #
         # @return nil
         # @api private
         def install_puppet_agent_from_dmg_on(hosts, opts)
-          opts[:puppet_collection] ||= 'PC1'
+          # TODO: this should default to nil, but we haven't enabled the rolling
+          # 'puppet' repos for mac yet. Remove the following line when those are
+          # available.
+          opts[:puppet_collection] ||= '5'
           opts[:puppet_collection] = opts[:puppet_collection].upcase #needs to be upcase, more lovely consistency
           block_on hosts do |host|
 
@@ -716,7 +714,13 @@ module Beaker
 
             variant, version, arch, codename = host['platform'].to_array
 
-            download_url = "#{opts[:mac_download_url]}/#{version}/#{opts[:puppet_collection]}/#{arch}"
+            # TODO: (Next Major Version) Remove the ability to install the
+            # pc1 package once puppet-agent 1.x has reached EOL
+            if opts[:puppet_collection] == 'PC1'
+              download_url = "#{opts[:mac_download_url]}/#{version}/#{opts[:puppet_collection]}/#{arch}"
+            else
+              download_url = "#{opts[:mac_download_url]}/puppet#{opts[:puppet_collection]}/#{version}/#{arch}"
+            end
 
             latest = get_latest_puppet_agent_build_from_url(download_url)
 
@@ -903,17 +907,20 @@ module Beaker
         alias_method :install_puppet_from_gem,          :install_puppet_from_gem_on
         alias_method :install_puppet_agent_from_gem_on, :install_puppet_from_gem_on
 
-        # Install official puppetlabs release repository configuration on host(s).
+        # Install official Puppet, Inc. release repository configuration on host(s).
         #
         # @param [Host, Array<Host>, String, Symbol] hosts    One or more hosts to act upon,
         #                            or a role (String or Symbol) that identifies one or more hosts.
+        # @param [String] repo    Optional, which major-version series puppet repos to enable. Only
+        #                           available for puppet5 and later. If left blank or set to nil, it
+        #                           will enable the latest by default.
         #
         # @note This method only works on redhat-like and debian-like hosts.
         #
         def install_puppetlabs_release_repo_on( hosts, repo = nil, opts = options )
           block_on hosts do |host|
             variant, version, arch, codename = host['platform'].to_array
-            repo_name = repo.nil? ? '' : '-' + repo
+            repo_name = repo.nil? ? '' : repo
             opts = FOSS_DEFAULT_DOWNLOAD_URLS.merge(opts)
 
             case variant
@@ -927,8 +934,15 @@ module Beaker
                 variant_url_value = 'cisco-wrlinux'
                 version = '7'
               end
-              remote = "%s/puppetlabs-release%s-%s-%s.noarch.rpm" %
-                [opts[:release_yum_repo_url], repo_name, variant_url_value, version]
+              # TODO: (Next Major Version) Remove the ability to install the
+              # pc1 release package once puppet-agent 1.x has reached EOL
+              if repo_name == 'pc1'
+                remote = "%s/puppetlabs-release-%s-%s-%s.noarch.rpm" %
+                  [opts[:release_yum_repo_url], repo_name, variant_url_value, version]
+              else
+                remote = "%s/puppet%s/puppet%s-release-%s-%s.noarch.rpm" %
+                  [opts[:release_yum_repo_url], repo_name, repo_name, variant_url_value, version]
+              end
 
               if variant == 'cisco_nexus'
                 # cisco nexus requires using yum to install the repo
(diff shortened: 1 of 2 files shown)


lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -438,11 +437,7 @@ def configure_puppet_on(hosts, opts = {})
         # @api private
         def install_puppet_from_rpm_on( hosts, opts )
           block_on hosts do |host|
-            if host[:type] == 'aio'
```
melissa: @kevpl how often are we not setting host type to aio? In the code prior to this change, that would enable the old style repos, which are no longer updated at all (we're trying to archive them at some point). How dangerous is it if I make this change?

kevpl: @melissa since we're no longer supporting Puppet 3.x, it seems like we should always be setting `host[:type]` to `aio`. Unfortunately, Beaker defaults to `pe` as its type, so there's definitely a chance that this won't be the case as people slip through the cracks. Our testing should help give us some confidence in this change 😺

kevpl:

lib/beaker-puppet/install_utils/foss_utils.rb
melissa: Should I keep the aio check in the case we have the host type set to PE?

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -918,8 +916,13 @@ def install_puppetlabs_release_repo_on( hosts, repo = nil, opts = options )
...
               end
-              remote = "%s/puppetlabs-release%s-%s-%s.noarch.rpm" %
-                [opts[:release_yum_repo_url], repo_name, variant_url_value, version]
+              if repo_name == 'pc1'
+                remote = "%s/puppetlabs-release-%s-%s-%s.noarch.rpm" %
+                  [opts[:release_yum_repo_url], repo_name, variant_url_value, version]
+              else
+                remote = "%s/puppet%s/puppet%s-release-%s-%s.noarch.rpm" %
+                  [opts[:release_yum_repo_url], repo_name, repo_name, variant_url_value, version]
+              end
```
melissa: I tried to keep this as backward compatible as possible, assuming that folks are no longer using the old puppetlabs repos that came before PC1. If people are still using those pre-pc1 repos, they are way way out of date

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -331,15 +331,14 @@ def install_puppet_on(hosts, opts={})
...
-        # @option opts [String] :puppet_collection Defaults to 'pc1'
+        # @option opts [String] :puppet_collection Defaults to nil
         # @option opts [Boolean] :run_in_parallel Whether to run on each host in parallel.
         #
         # @return nil
         # @raise [StandardError] When encountering an unsupported platform by default, or if gem cannot be found when default_action => 'gem_install'
         # @raise [FailTest] When error occurs during the actual installation process
         def install_puppet_agent_on(hosts, opts = {})
           opts = FOSS_DEFAULT_DOWNLOAD_URLS.merge(opts)
-          opts[:puppet_collection] ||= 'pc1' #hi!  i'm case sensitive!  be careful!
```
melissa: So we're just not setting a default value here, but folks can still pass this value in. If it's set to `pc1` (case sensitive), it'll do the same as previous, which is enable the pc1 repos. If left blank, it'll default to the puppet rolling repos. Other than pc1, possible values are going to correspond to any major version series. Right now the only one out is `5`, but once puppet6 is live, it could be set to `6`, etc.

spec/beaker-puppet/install_utils/foss_utils_spec.rb
```diff
@@ -448,15 +448,15 @@ def logger
     context 'on el-6' do
       let(:platform) { Beaker::Platform.new('el-6-i386') }
       it 'installs' do
-        expect(hosts[0]).to receive(:install_package_with_rpm).with(/puppetlabs-release-el-6\.noarch\.rpm/, '--replacepkgs', {:package_proxy=>false})
+        expect(hosts[0]).to receive(:install_package_with_rpm).with(/puppet-release-el-6\.noarch\.rpm/, '--replacepkgs', {:package_proxy=>false})
```
melissa: I should probably add more tests to verify this functionality is properly handled

lib/beaker-puppet/install_utils/foss_utils.rb
kevpl: @melissa so if I'm understanding this right, if people don't provide a `repo_name`, an example url beaker could generate is:
```
release_yum_repo_url/puppet/puppet-release-el-6.noarch.rpm
```
but to get puppet 5 or 6, they provide that as the `repo_name`, and the url example above would be:
```
release_yum_repo_url/puppet5/puppet5-release-el-6.noarch.arm
```
Is this correct?

Looking at how you've added in the backwards compatible case, it looks like the strings don't match up:
```
old: %s/puppetlabs-release%s-%s-%s.noarch.rpm
new: %s/puppetlabs-release-%s-%s-%s.noarch.rpm
```
Was the old one incorrect? Once we get that discrepancy settled, I'm down with leaving this as a strategy for backwards compatibility. If you would make a comment for us to try to make sure to come back & get rid of this case in the future, that'd be cool ([example](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host/pswindows/exec.rb#L183)).

spec/beaker-puppet/install_utils/foss_utils_spec.rb
kevpl: I'm not gonna say no to more accurate / specific test coverage 🙈

kevpl:
@melissa it looks like spec testing did fail on these changes, so that will need looking at

lib/beaker-puppet/install_utils/foss_utils.rb
melissa: on line 907 that I changed, it was previously adding in the `-`. This code assumes users will never want to enable the old `puppetlabs-release-(<platform>-<verision>|<codename>)` packages, so there isn't a case where we need to guard that dash before the repo name. So I figured this would be a little more intuitive. So, that's one caveat to preserving backward compatibility. Users will no longer be able to access those old release packages. Instead, it will default to the newest release packages.

And yes, `release_yum_repo_url/puppet/puppet-release-el-6.noarch.rpm` and `release_yum_repo_url/puppet5/puppet5-release-el-6.noarch.arm` are correct strings. Did you see my email to users explaining the difference between these two repos? You can find it at https://groups.google.com/d/msg/puppet-announce/tKHAKlWnXFA/_Mdsoxo4BAAJ. I didn't send it out internally, though now I'm reconsidering that

melissa: Also, yes, I'm happy to add a note to remove that case in the future

melissa:
Dang! I thought I had already fixed the tests. I'll fix that when I add and update the existing tests around this code

spec/beaker-puppet/install_utils/foss_utils_spec.rb
```diff
@@ -668,14 +662,35 @@ def logger
...
       end
 
     end
 
+    describe "when puppet_collection is 'pc1'" do
+      describe "on yum-based platforms" do
+        let( :platform ) { Beaker::Platform.new('el-7-x86_64') }
+
+        it "installs the old style puppetlabs-release-pc1 packages" do
+          expect(host).to receive(:execute).with( /rpm .*puppetlabs-release-pc1.*\.rpm$/ )
```
melissa: @kevpl could I get some help with this? I'm not great at rspec, but this looks like it should be correct, except I'm getting the following error:

```
  1) ClassMixedWithDSLInstallUtils#install_puppetlabs_release_repo when puppet_collection is 'pc1' on yum-based platforms installs the old style puppetlabs-release-pc1 packages
     Failure/Error:
       host.install_package_with_rpm( remote, '--replacepkgs',
         { :package_proxy => opts[:package_proxy] } )

       #<Unix::Host:0x007fde520f9d70 @logger=#<Double "logger">, @options={:logger=>#<Double "logger">}, @host_hash={:user=>"root", :group=>"root", :pathseparator=>":", :platform=>"el-7-x86_64"}, @name="fak
evm", @apt_needs_update=true, @external_copy_base=nil, @commands=[]> received :execute with unexpected arguments
         expected: (/puppetlabs-release-pc1.*\.rpm$/)
              got: ("rpm --replacepkgs -Uvh http://yum.puppetlabs.com/puppetlabs-release-pc1-el-7.noarch.rpm ")
       Diff:
       @@ -1,2 +1,2 @@
       -[/puppetlabs-release-pc1.*\.rpm$/]
       +["rpm --replacepkgs -Uvh http://yum.puppetlabs.com/puppetlabs-release-pc1-el-7.noarch.rpm "]
```

melissa: NEVERMIND! It's the space at the end of the line

melissa:
@kevpl Updated and ready for another review!

lib/beaker-puppet/install_utils/foss_utils.rb
kevpl: @melissa if I'm understanding your changes right, we're moving to a situation where we can have `aio` runs where there are multiple different artifacts that could be installed, right? Typically those would be `pc1` for Puppet 4 artifacts, `nil` for the latest Puppet artifacts, or specific version numbers if those are wanted specifically in the new release structure, is that right?

If that's the case, we can see from [this code](https://github.com/puppetlabs/beaker/commit/688a1c8b4042723fb578f2ded19e63c8e0c32ed4) that people have been using this method to install `pc1` artifacts. The change you've proposed takes this ability away from them, as there's no way to pass it in anymore. I'm OK with de-coupling the `aio` run type from the `repo_name` parameter, but we'll have to provide some ability to still pass `pc1` to the `install_puppetlabs_release_repo` method from this one. I'm fine with an `opts[:puppet_collection]` parameter like we've used elsewhere (even though we're including things that aren't collections anymore (😞 in naming)).

melissa: Ah yes, I see where you're coming from here. That makes sense. Yes, I'll add in a work around for this. Do you think it's safe to change the default repo that is enabled, or would that be too much of a breaking change?

kevpl: Yeah, I think that's fine as long as we have a way to install the older stuff as well.

melissa:
@kevpl updated! I also found a few additional places that looked like they ought to be updated for puppet5 land

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -438,11 +437,7 @@ def configure_puppet_on(hosts, opts = {})
         # @api private
         def install_puppet_from_rpm_on( hosts, opts )
           block_on hosts do |host|
-            if host[:type] == 'aio'
-              install_puppetlabs_release_repo(host,'pc1',opts)
-            else
-              install_puppetlabs_release_repo(host,nil,opts)
-            end
+            install_puppetlabs_release_repo(host,opts[:puppet_collection],opts)
```
melissa: @kevpl is this reasonable to have folks pass in `puppet_collection` to opts if they want to use something other than the default repo?

melissa:
Updated!

lib/beaker-puppet/install_utils/foss_utils.rb
melissa: When you say the "older stuff" are you talking about the pc1 repos, or are you talking about the old old repos? I.e., on yum, we have the `products` repos that we used to ship to http://yum.puppetlabs.com/el/6/products/

kevpl: I mean the older things that it could install before these changes. Which, looking at this code path in the method, appear to be just the pc1 repos rather than that & the old `products` repos.

lib/beaker-puppet/install_utils/foss_utils.rb
kevpl: it seems that they'll have to pass something in to use anything other than the default repos. Are you thinking that the name should be different, that they shouldn't be able to specify something other than the default repo, or something else?

melissa:
Closing in favor of https://github.com/puppetlabs/beaker-puppet/pull/18

PR #12, opened 2017-08-02
(MAINT) fix windows test failures

kevpl:
spec tests recently started failing here on Windows, I believe due to recent windows changes upstream in beaker. This PR fixes those tests.

The changes upstream weren't detected because current beaker testing runs beaker-puppet's acceptance tests, not the spec ones. We should talk about perhaps using beaker-puppet's Combined Smoketest if possible (the same ones here) rather than just the one.


kevpl:
Closing in favor of #14

PR #13, opened 2017-08-04
add proxy for windows agent install

tomqwu:


tomqwu:
Anyone?

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -602,7 +602,7 @@ def install_a_puppet_msi_on(hosts, opts)
 
             if host.is_cygwin?
               # NOTE: it is critical that -o be before -O on Windows
-              on host, "curl -o \"#{msi_download_path}\" -O #{link}"
+              on host, "export http_proxy=#{opts[:package_proxy]}  && curl -o \"#{msi_download_path}\" -O #{link}"
```
ferglor: Would it not be better to use the `-x` or `--proxy` options in curl, rather than exporting an environment variable? Also I'm not sure if http_proxy accepts a proxy that uses the https protocol, as I believe there is a separate https_proxy env var, specifically for https protocols. For that reason, I'd be inclined to use the `-x` or `--proxy` flags in curl, which would allow the user to specify a proxy using both the http and https protocols

tomqwu: If I am using `-x`, then every curl will use the proxy, what if the proxy is empty?

ferglor: In that case I would be inclined to initialise a proxy option var, something like: `proxy = opts[:package_proxy] ? "-x #{opts[:package_proxy]}" : ""` and then use that in your curl command the same way you're using msi_download_path. That being said, it is a little cumbersome. My main concern is how we handle the scenario where a HTTPS proxy is specified in opts.

tomqwu: this makes sense, I will update it accordingly

tomqwu:
@ferglor please review the change again, thanks

ferglor:
@tomqwu this looks good, can you update the spec tests, or add new spec tests specifically for this change.Thanks!

mchllweeks:
last comment is current

kevpl:
last comment is still current. If there are no changes in a week, we'll close this PR due to inactivity.

kevpl:
closed due to inactivity

lib/beaker-puppet/install_utils/foss_utils.rb
lmayorga1980: I believe there is no support for a windows environment to use a proxy like for instance the opentable vagrant boxes. Are there any plans to add support for a proxy?

PR #14, opened 2017-08-04
(MAINT) fix windows spec failures

tvpartytonight:
Changes in beaker core resulted in failures here; these changes update
the spec tests to use an actual Beaker::Platform object instead of a
String object.


tvpartytonight:
@kevpl these are the changes that we discussed to do instead of #12 .

PR #15, opened 2017-08-10
(BKR-1185) Use Oga instead of Nokogiri

rishijavia:
1. beaker-puppet did not have an explicit nokogiri dependency declaration as it
was using the one provided with beaker-core. Beaker-core is will not have
nokogiri as a dependency soon. Therefore adding oga as a dependency for
beaker-puppet.

2. The old get_latest_puppet_agent_build_from_url method did not return the
right version number even on the url provided in the documentation of the
method. The only url it returned a correct verion was:
"https://downloads.puppetlabs.com/mac/10.9/PC1/x86_64"
With the new code some of the URLs that it parses correctly include:
"https://downloads.puppetlabs.com/mac/10.9/PC1/x86_64"
"https://downloads.puppetlabs.com/mac/10.10/PC1/x86_64"
"https://downloads.puppetlabs.com/mac/10.11/PC1/x86_64"
"https://downloads.puppetlabs.com/mac/10.12/PC1/x86_64"
"https://downloads.puppetlabs.com/windows"

### DIFF
--- a/lib/beaker-puppet/install_utils/foss_utils.rb
+++ b/lib/beaker-puppet/install_utils/foss_utils.rb
@@ -744,17 +744,26 @@ module Beaker
         #                             Empty string if none found.
         # @api private
         def get_latest_puppet_agent_build_from_url(url)
-          require 'nokogiri'
-          require 'open-uri'
-          page = Nokogiri::HTML(open("#{url}/?C=M;O=A"))
-          agents = page.css('a').children.select{ |link| link.to_s.include? 'puppet-agent' }
-          re =  /puppet-agent-(.*)-1/
-          latest_match = agents[-1].to_s.match re
+          require 'oga'
+          require 'net/http'
+          body = Net::HTTP.get(URI.parse("#{url}/index_by_lastModified_reverse.html"))
+          document = Oga.parse_html(body)
+          agents = document.xpath('//a[contains(@href, "puppet-agent")]')
+
+          latest_match = agents.shift.attributes[0].value
+          while (latest_match =~ /puppet-agent-\d(.*)/).nil?
+            latest_match = agents.shift.attributes[0].value
+          end
+
+          re  =  /puppet-agent-(.*)-/
+          latest_match = latest_match.match re
+
           if latest_match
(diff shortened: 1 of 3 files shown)


kevpl:
@johnduarte I believe you added this method originally, is that right? We'd love a review if you have some time.

kevpl:
in some local trials, I'm seeing a speedup from a little over a minute for nokogiri to install to ~2.5 seconds for oga, so that seems good enough for me to approve of it 👍

johnduarte:
My suggestion above does not work, because the dmg package contains the version number and will not install using 'latest' as a version.

rishijavia:
@johnduarte so we still need the `get_latest_puppet_agent_build_from_url` method?

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -744,17 +744,26 @@ def install_puppet_agent_from_dmg_on(hosts, opts)
...
         # @api private
         def get_latest_puppet_agent_build_from_url(url)
-          require 'nokogiri'
-          require 'open-uri'
-          page = Nokogiri::HTML(open("#{url}/?C=M;O=A"))
-          agents = page.css('a').children.select{ |link| link.to_s.include? 'puppet-agent' }
-          re =  /puppet-agent-(.*)-1/
-          latest_match = agents[-1].to_s.match re
+          require 'oga'
+          require 'net-http'
```
johnduarte: This fails to load for me. Changing this to the stdlib 'net/http' as documented in [stdlib](http://ruby-doc.org/stdlib-2.4.1/libdoc/net/http/rdoc/Net/HTTP.html) allows this to successfully load. Please change this to the following.

```
require 'net/http'
```

johnduarte:
@rishijavia yes, we still need the `get_latest_puppet_agent_build_from_url` method. :-(. Modify the `net/http` require and your PR will get my approval. Thank you for your efforts in expunging nokogiri!

rishijavia:
@kevpl, @johnduarte PR updated :+1:

PR #20, opened 2017-10-09
Updates to support GCE and RHEL

trevor-vaughan:
Added RHEL support across the baseline

### DIFF
--- a/lib/beaker-puppet/version.rb
+++ b/lib/beaker-puppet/version.rb
@@ -1,3 +1,3 @@
 module BeakerPuppet
-  VERSION = '0.8.0'
+  VERSION = '0.7.1'
 end
--- a/lib/beaker-puppet/version.rb
+++ b/lib/beaker-puppet/version.rb
@@ -1,3 +1,3 @@
 module BeakerPuppet
-  VERSION = '0.7.1'
+  VERSION = '0.7.0'
 end
--- a/lib/beaker-puppet/helpers/puppet_helpers.rb
+++ b/lib/beaker-puppet/helpers/puppet_helpers.rb
@@ -291,7 +291,7 @@ module Beaker
         # @!visibility private
         def dump_puppet_log(host)
           syslogfile = case host['platform']
-            when /fedora|centos|el|redhat|scientific/ then '/var/log/messages'
+            when /fedora|centos|el|rhel|redhat|scientific/ then '/var/log/messages'
             when /ubuntu|debian|cumulus/ then '/var/log/syslog'
             else return
           end
--- a/lib/beaker-puppet/install_utils/ezbake_utils.rb
+++ b/lib/beaker-puppet/install_utils/ezbake_utils.rb
@@ -69,7 +69,7 @@ module Beaker
         # @api private
         def ezbake_validate_support host
           variant, version, _, _ = host['platform'].to_array
-          unless variant =~ /^(fedora|el|centos|debian|ubuntu)$/
+          unless variant =~ /^(fedora|el|rhel|redhat|centos|debian|ubuntu)$/
             raise RuntimeError,
                   "No support for #{variant} within ezbake_utils ..."
           end
--- a/lib/beaker-puppet/install_utils/foss_utils.rb
+++ b/lib/beaker-puppet/install_utils/foss_utils.rb
@@ -347,7 +347,7 @@ module Beaker
             add_role(host, 'aio') #we are installing agent, so we want aio role
             package_name = nil
             case host['platform']
-            when /el-|fedora|sles|centos|cisco_/
+            when /el-|rhel|redhat|fedora|sles|centos|cisco_/
               package_name = 'puppet-agent'
               package_name << "-#{opts[:puppet_agent_version]}" if opts[:puppet_agent_version]
             when /debian|ubuntu|cumulus|huaweios/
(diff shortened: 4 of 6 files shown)


lib/beaker-puppet/version.rb
```diff
@@ -1,3 +1,3 @@
 module BeakerPuppet
-  VERSION = '0.7.0'
+  VERSION = '0.7.1'
```
kevpl: please pull this out of the changes

kevpl:
last comment still current

trevor-vaughan:
@kevpl I thought I explained this but I have no idea where it got tossed.

Different images in GCE, and other systems, can show up as either `rhel` or `redhat` which is why both were added.

kevpl:
@trevor-vaughan why would that matter to beaker? Wouldn't beaker treat them both as just redhat platforms & you could make that distinction elsewhere, like in a google hypervisor-specific property?

trevor-vaughan:
@kevpl By that logic, `amazon` shouldn't be in there because it's literally a `redhat` clone and no additional work is required internal to beaker.

You don't want to make the user do extra steps for something that can be done transparently by the application and, in this case, besides looking a bit fugly, you're going to make the user do extra work if you don't include it in the code.

kevpl:
@trevor-vaughan I agree that `amazon` shouldn't be there, and we're hoping to pull that out at some point in the future.

The idea of doing something transparently in beaker is nice, but it never turns out to be as nice as we thought when people break down the road & every bit of magic we hid from them turns into a new place for the tests to break at & cause failures. Our newer policy is to have beaker be as simple as possible & make users pass information that's needed to do something rather than try to do a little more to default to something that works, but results in something that's harder to understand & debug.

Platform strings are a particularly bad thing to override & make "just work," as our historic pattern for changing functionality based on platform is a bunch of custom snowflake case statements. In particular, adding platforms that have the same behavior as other platforms is bad because there's always another place where the case statement needs to be updated to include another copied platform. Unfortunately the people who find those cases tend to be the end user, not the platform implementer, & the output that they receive doesn't tend to help point them to this issue as the cause of their problem either.

With this taken into account, we'd prefer to have to make a beaker user map their redhat variants to a redhat platform string when they want to run beaker tests rather than have beaker accept a bunch of string variants & try to juggle them.

If you'd like to take on making beaker more open to other platform variants, but have them map to one platform during beaker's startup so that we only complicate the parsing essentially for something like this, that's something we'd be more open to merging & could allow you to have `rhel` platform strings you're looking for. Unfortunately, that work isn't planned internally in the near future.

trevor-vaughan:
@kevpl Ok, so **once again** you're changing the rules on me mid-stride and it's getting *very* irritating.

What I'm taking away from this:

1.  It's OK for the `beaker` core team to do bad things and push them through, but not contributors
2. You don't want to fix this problem properly later, you want to have `amazon` work easily and not `gce` (no idea why since it would be the same fix later)

Now, I would like to suggest that we go ahead with what works for GCE now and fix the overall issue as a separate issue at a later date (which would include `amazon`, `el`, `rhel`, and everything else) *or* we make everything even and rip out `amazon` at the same time.

Alternatively, we can abandon the entire GCE stack, or you can mangle it yourself and add an explanation as to whatever insanity users need to do to make things work, I've got 'Allow maintainers to update' on my branches so you can make it as difficult as you like. I simply don't have the time to jump through absolutely useless hoops as well as figure out and document all of the edge cases and procedures when adding one word to an already hideous regex works just fine and literally has an example of doing just that that was pushed **while I was updating the GCE PR to work with the latest Beaker because it was ignored for a LONG time earlier**.

kevpl:
@trevor-vaughan, we are definitely changing the rules all the time, I can understand how frustrating that can be, especially to just see the outputs of those, rather than to see the processes as well.

In the `amazon` case, in particular, I consider it a moment of weakness that this was merged to begin with. It's not very surprising that these happen, PR review is a human process, mistakes will be made. Fortunately, the use-case was so specific & focused that we fairly quickly saw that this was a bad idea. Thanks to the conversation we're having here bringing the topic up again, it gave us time within the "maintainer" budget to get that revert going (https://github.com/puppetlabs/beaker/pull/1471). We'd seen the error of our ways, but not gotten the time budget to spend fixing this until we had to have more conversation on the topic, so thank you for bringing that up.

In terms of your paragraph beginning "Alternatively," it has been our experience that simplifying the platforms rather than letting them umbrella out into a bunch of variants has resulted in simpler code & documentation. Unfortunately, we don't have the time budget allocated to spend adding things to most PRs. Unless the PR is fairly small, requiring us to add things will result in an unchanged & closed PR.

kevpl:
@trevor-vaughan I'm OK with adding the redhat value, and most of your work to get GCE support going again. But beaker should not need to manage a particular hypervisor's quirks (in this case that GCE returns both values) across most of its code.

The specific hypervisor should act as an adapter to the quirks, translating between both systems, so that someone who doesn't use GCE (for instance) doesn't have to know this quirk to implement a new feature or fix a bug in beaker itself. If there's some reason that the hypervisor can't act as an adapter in this fashion, let me know why that is, and we'll discuss other options then.

With things as they are, however, if these PRs aren't moved forward in a month, we'll be closing them due to inactivity.

kevpl:
@trevor-vaughan last message is still current. If no progress is made on this PR in a week, it will be closed due to inactivity.

trevor-vaughan:
@kevpl Request to reopen based on https://github.com/puppetlabs/beaker-puppet/pull/27

PR #21, opened 2017-11-28
(maint) Add beaker-abs to Gemfile

smcelmurry:
This is needed for acceptance testing in production. It should not have an effect on local testing.


kevpl:

PR #23, opened 2018-01-04
(BKR-1343) Install as an MSI on Windows

joshcooper:
On Windows, the "install_from_build_data_url" method was trying to
install the artifact as a cygwin package which won't work for MSIs.
Instead use the beaker method for installing as an MSI. Note the
"install_msi_on" method is specific to the puppet-agent package, because
it will ensure the puppet service is stopped before installing and will
install it as a Manual service. In the future we may want to
differentiate between the puppet-agent project and others, and use
"generic_install_msi_on" for the latter.

### DIFF
--- a/lib/beaker-puppet/install_utils/puppet5.rb
+++ b/lib/beaker-puppet/install_utils/puppet5.rb
@@ -102,7 +102,9 @@ module BeakerPuppet
           onhost_package_file = "#{ project_name }*"
         end
 
-        if variant == 'eos'
+        if variant == 'windows'
+          install_msi_on(host, artifact_url)
+        elsif variant == 'eos'
           # TODO Will be refactored into {Beaker::Host#install_local_package}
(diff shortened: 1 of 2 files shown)


joshcooper:
Relates to https://github.com/puppetlabs/puppet/pull/6477

melissa:
I'm concerned about AIX here as well. Currently it looks like `install_local_package` doesn't have aix support at all yet. That could definitely wait until we get aix issues in the yaml file resolved though

joshcooper:

melissa:

melissa:
@kevpl any thoughts on why tests aren't getting kicked off? We'd like to get this merged in today if possible

tvpartytonight:
@melissa I think no runs are starting because it only starts runs for whitelisted github users

tvpartytonight:

tvpartytonight:

PR #24, opened 2018-01-05
(maint) Refactor puppet5 install logic

joshcooper:
This is blocked on https://github.com/puppetlabs/beaker-puppet/pull/23

Previously the if/elsif block set onhost_package_file for some, but not
all hosts. Then a second if/elsif block called different install methods
based on the platform variant and the above variable.

This commit uses a single case/when statement to branch. It makes it
clearer that EOS calls install_from_file, OSX uses the single argument
version of install_local_package, while Solaris uses the two argument
version. Also Windows uses a different installation method entirely.


kevpl:
👍 will kick testing once #23 is merged, then we should be good to merge & release

kevpl:

PR #26, opened 2018-02-02
(BKR-1342) Import ci rake tasks and install utils

melissa:
This commit moves all ci acceptance rake tasks and install/setup
utilities out of the puppet repo and into the beaker-puppet repo. This
will allow us to reuse all of this code across multiple components of
puppet-agent. The goal is to reduce the number of places we need to
duplicate code across different code bases. With this change, any future
updates or overhauls will be consistent across all our projects.


melissa:
This is still blocked on a few updated to puppet, so it's not yet ready to be merged

melissa:
This PR is based on changes made in https://github.com/puppetlabs/puppet/pull/6550

lib/beaker-puppet/helpers/setup/common.rb
```diff
@@ -0,0 +1,187 @@
...
+                  end
+                else
+                  raise("Unknown platform '#{platform_key}' in package_hash")
+                end
+              end
+            end
+            return true
+          end
+
+          def ruby_command(host)
```
melissa: I don't really like these here, but I was unsure where to drop them. They are utilities that are only used by tests defined in `common.rb`, `gem.rb`, `aio.rb`, etc. They are not referenced by a component when running tests, as are the rest of the methods in this file. I think it would be good to move these to a different file, but I wasn't sure the best place for that

melissa:
I also want to note that I did not bring along any of the pe-specific utilities, specifically these ones I left out:
https://github.com/puppetlabs/puppet/blob/44e1a85194583bd350c8a79c3c4e0494fc4b144f/acceptance/setup/common/pre-suite/110_SetPEPuppetService.rb#L1

https://github.com/puppetlabs/puppet/blob/44e1a85194583bd350c8a79c3c4e0494fc4b144f/acceptance/setup/pe/pre-suite/000_Install.rb#L1

https://github.com/puppetlabs/puppet/blob/44e1a85194583bd350c8a79c3c4e0494fc4b144f/acceptance/setup/pe/pre-suite/010_UpdatePkg.rb#L1

melissa:
@kevpl I just realized you already have a lot of this code in the `acceptance` directory of beaker-puppet (https://github.com/puppetlabs/beaker-puppet/tree/master/acceptance). Was that intended to be used by projects like puppet and hiera to run acceptance tests?

lib/beaker-puppet/helpers/setup/aio.rb
```diff
@@ -0,0 +1,112 @@
...
+              end
+
+              configure_gem_mirror(hosts)
+            end
+          end
+
+          def install_cumulus_modules
+            test_name "Install Cumulus Modules" do
+              platforms = hosts.map{|val| val[:platform]}
+              skip_test "No cumulus hosts present" unless platforms.any? { |val| /cumulus/ =~ val }
```
melissa: @joshcooper it looks like this skip is causing the top level test (https://github.com/puppetlabs/puppet/pull/6602/files#diff-62e97df027f65c15385cb6a34498e2acR1) to skip, unless I'm missing something? Is this something that happens with nested tests?

```
  Install Cumulus Modules

    No cumulus hosts present
Begin teardown
End teardown
Warning: setup/aio.rb skipped in 91.94 seconds
      Test Suite: pre_suite @ 2018-02-06 11:10:42 -0800

      - Host Configuration Summary -


              - Test Case Summary for suite 'pre_suite' -
       Total Suite Time: 91.94 seconds
      Average Test Time: 91.94 seconds
              Attempted: 1
                 Passed: 0
                 Failed: 0
                Errored: 0
                Skipped: 1
                Pending: 0
                  Total: 1

      - Specific Test Case Status -

Failed Tests Cases:
Errored Tests Cases:
Skipped Tests Cases:
  Test Case setup/aio.rb
Pending Tests Cases:

No tests to run for suite 'tests'
No tests to run for suite 'post_suite'
No tests to run for suite 'pre_cleanup'
Cleanup: cleaning up after successful run
```

melissa: alright, I got this figured out

kevpl:
@melissa to this question:

    @kevpl I just realized you already have a lot of this code in
    the acceptance directory of beaker-puppet
    (https://github.com/puppetlabs/beaker-puppet/tree/master/acceptance).
    Was that intended to be used by projects like puppet and hiera to
    run acceptance tests?

No, it wasn't, it was put there just to help beaker-puppet do its own acceptance testing. However, we'd much rather have a "true" (meaning not in beaker-puppet's acceptance helpers) DSL method that both you & we can use in testing. So my recommendation would be that we switch beaker-puppet's testing to use your versions of the methods & delete beaker-puppet-acceptance's copy of them.

I understand that this is additional work for what is already potentially a big glob of work, so I don't think this needs to necessarily be done before this PR is merged. If you'd like, we can ticket this work for BKR & hopefully be able to schedule this soon with a goalie. @melissa would you like to create the ticket since your head is in this right now, or should I do that?

melissa:
@kevpl created https://tickets.puppetlabs.com/browse/BKR-1424 for that work. Thank you!

Also, this is blocked because of a bug we found where beaker subcommands don't work well when passing in additional options. BKR-1422 was created to cover this particular bug.

tasks/ci.rake
```diff
@@ -0,0 +1,282 @@
...
+  sh('rspec lib')
+end
+
+def beaker(command, *argv)
+  argv.concat(ENV['OPTIONS'].split(' ')) if ENV['OPTIONS']
+
+  sh('beaker', command.to_s, *argv)
+end
+
+def beaker_suite(type)
```
melissa: @joshcooper this is what I have so far. Not necessarily a final version, but at least it's proof this is working!

lib/beaker-puppet/helpers/rake_helpers.rb
```diff
@@ -0,0 +1,21 @@
+module Beaker
+  module DSL
+    module Helpers
+      # Methods that help you interact with rake during ci setup
+      module RakeHelpers
+        class << self
+          def load_tasks(beaker_root = '/Users/melissa/beaker-puppet')
```
joshcooper: can we make it a required parameter?

melissa: oh shoot, I meant to fix this, but I totally forgot. Thank you! I don't think we need to make it required, since we should be able to introspect where beaker-puppet is being run. I do that for the setup tests

lib/beaker-puppet/helpers/rake_helpers.rb
```diff
@@ -0,0 +1,21 @@
+module Beaker
+  module DSL
+    module Helpers
+      # Methods that help you interact with rake during ci setup
+      module RakeHelpers
+        class << self
+          def load_tasks(beaker_root = File.dirname(__dir__))
```
melissa: @joshcooper fixed!

melissa:
I tested aio, gem, and git. Aio and gem both succeeded. Git failed, but in an expected way and should not be a blocker.

tasks/ci.rake
```diff
@@ -0,0 +1,283 @@
...
+basis.
+
+  $ SHA=<full sha> bundle exec rake ci:test:quick
+
+SHA should be the full SHA for the puppet-agent package.
+EOS
+    task :quick => ['ci:check_env', 'ci:gen_hosts'] do
+      ENV['TESTS'] = get_test_sample.join(",")
+      Rake::Task["ci:test:aio"].invoke
+    end
```
joshcooper: The `ci:test:quick` task is specific to the puppet component, e.g. the `get_test_sample` method returns tests that only live in puppet. How about keeping this task (and related helper methods) in puppet's Rakefile?

melissa:
@joshcooper I removed the `ci:test:quick` task and associated utilities

@ericldelaney did you want me to add a check for if `--preserve-hosts=always` is present in OPTIONS and skip `beaker destroy` if it is?

ericldelaney:
@melissa it would be nice if at some point we honored --preserve-hosts correctly. It would save me from having to create a failing test to run against so that it leaves the machines. (shrug)

melissa:
ticket filed https://tickets.puppetlabs.com/browse/BKR-1443 feel free to add more detail

PR #27, opened 2018-02-26
(MAINT) Updates to support GCE and RHEL

trevor-vaughan:
Added RHEL support across the baseline


kevpl:
Hmm, looks like it didn't work in the review comment. I'll try again here..

Jenkins, test this please

kevpl:
I believe #69 will get merged soon, so I'm planning on delaying releasing this until that one merges as well. If it doesn't move soon though, I'll cut a new beaker-puppet regardless.

PR #28, opened 2018-03-16
(BKR-1442) do not reuse opts across multiple hosts

DavidS:


mchllweeks:
@DavidS Could you please add a test to ensure we do not break this in the future.

DavidS:
No.

DavidS:
PS: for my use-case I've worked around it by installing modules to each host individually: https://github.com/puppetlabs/puppetlabs-resource_api/blob/527f7c0b8dce8dcc4da191d6020bb90aa2973973/spec/spec_helper_acceptance.rb#L14-L15

kevpl:
per an out of band discussion, we've decided to close this PR for now & leave the bug open for further work to fix this issue if it comes up for others.

PR #29, opened 2018-03-20
(maint) Remove references to 'ci:test:quick'

melissa:
We decided to isolate this test to the puppet project. The tests that it
pulls are very specific to puppet itself, so we'd rather implement this
on a per project basis. I forgot to pull out the references to the quick


melissa:
@joshcooper @kevpl I forgot to pull out these references. This is completely cosmetic, and doesn't warrant a new release, but I want to get this in before I forget about it.

PR #30, opened 2018-03-22
(PA-1915) Test components against nightly puppetserver

melissa:
The layout of nightlies.delivery.puppetlabs.net recently changed, and we
were no longer able to use it for agent component testing. This commit
updates where we look for puppetserver when we want to pull from
nightlies.

This commit also breaks out agent installation, server installation, and
general install finalization into multiple tests. This will allow us to,
for example, skip installing puppetserver when running hiera tests.


setup/aio/011_Install_Puppet_Server.rb
```diff
@@ -0,0 +1,29 @@
+extend BeakerPuppet::Install::Puppet5
+extend Beaker::DSL::InstallUtils::FOSSUtils
+
+test_name "Install Puppet Server" do
+  server_version = ENV['SERVER_VERSION'] || 'latest'
+  release_stream = ENV['RELEASE_STREAM'] || 'puppet'
+  nightly_builds_url = ENV['NIGHTLY_BUILDS_URL'] || 'http://nightlies.puppetlabs.com'
```
mwaggett: This will work, but I think we want to prefer nightlies.puppet.com rather than puppetlabs.

mwaggett:
This seems reasonable to me! Can't speak to the gritty details of beaker, but this'll work afaict. 
Have you tested?

setup/aio/011_Install_Puppet_Server.rb
```diff
@@ -0,0 +1,29 @@
+extend BeakerPuppet::Install::Puppet5
+extend Beaker::DSL::InstallUtils::FOSSUtils
+
+test_name "Install Puppet Server" do
+  server_version = ENV['SERVER_VERSION'] || 'latest'
+  release_stream = ENV['RELEASE_STREAM'] || 'puppet'
+  nightly_builds_url = ENV['NIGHTLY_BUILDS_URL'] || 'http://nightlies.puppet.com'
```
melissa: @mwaggett updated

melissa:
Yes, I tested this.

mwaggett:
Since I don't know beaker stuff well, I don't want to give an official 👍, but this looks good as far as I can tell. 😊

tasks/ci.rake
```diff
@@ -201,7 +201,9 @@ def pre_suites(type)
   when :aio
     [
       "#{beaker_root}/setup/common/000-delete-puppet-when-none.rb",
-      "#{beaker_root}/setup/aio/010_Install.rb",
+      "#{beaker_root}/setup/aio/010_Install_Puppet_Agent.rb",
+      "#{beaker_root}/setup/aio/011_Install_Puppet_server.rb",
```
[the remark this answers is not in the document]
melissa: Interesting. This is indeed a typo, but somehow it still worked when I tested it. I'm pushing a fix up now.

setup/aio/011_Install_Puppet_Server.rb
```diff
@@ -0,0 +1,29 @@
+extend BeakerPuppet::Install::Puppet5
+extend Beaker::DSL::InstallUtils::FOSSUtils
+
+test_name "Install Puppet Server" do
+  server_version = ENV['SERVER_VERSION'] || 'latest'
+  release_stream = ENV['RELEASE_STREAM'] || 'puppet'
```
Magisus: Where is `RELEASE_STREAM` coming from? `build_default.yaml`? (asking because I was told to update that to point to 'puppet6' for our master branch, which seems like what we want here)

melissa: it'll have to be set in ci-job-configs, and it will correspond to which nightlies repo we hit. Right now, it can be one of `puppet`, `puppet5`, or `puppet6`. Since it's different based on the branch, we can't really set it here. Right now I think `puppet` links to `puppet6`, if I'm remembering correctly

mwaggett: `puppet-nightly` links to `puppet6-nightly`, yes

Magisus: 👍

joshcooper: Unfortunately the release package has the version in it, so the default `puppet` will try to download `http://nightlies.puppet.com/yum/puppet-nightly/puppet-nightly-release-el-7.noarch.rpm`, but fail because the release package is `http://nightlies.puppet.com/yum/puppet-nightly/puppet6-nightly-release-el-7.noarch.rpm`. How about if we default the `RELEASE_STREAM` to `puppet6` so that it works by default, and if an component branch needs an earlier nightlies version or released version they can set `RELEASE_STREAM` or `SERVER_VERSION` is ci-job-configs?

mwaggett: There are puppet-nightly-release packages on the top level (e.g. http://nightlies.puppet.com/yum/puppet-nightly-release-el-7.noarch.rpm) if that helps.

melissa: @mwaggett that doesn't help. Could you add symlinks for the release packages like you do in http://yum.puppetlabs.com/puppet/ ?

mwaggett: @melissa sure.

It seems generally fragile though to expect this specific layout though. How hard would it be to update the `install_puppetlabs_release_repo_on` method to accommodate whatever path?

mwaggett: fair. I agree that we should be consistent.

mwaggett: ooooh alternatively, you could just grab the nightly release packages that live on yum.puppetlabs.com and apt.puppetlabs.com. Those have the <stream>-nightly/<stream>-nightly-release structure you're expecting and are symlinks to the ones on nightlies.puppet.com.

melissa: Will those be updated for puppet7, or will those be just as fragile as updating the release packages on nightlies? Which ever we hit, I still think we need to get yum.pl.com and nightlies.pl.com/yum into a consistent state.

mwaggett: yes, I will work on getting things consistent today.

I'm not sure what you're asking in your first question though. We plan to keep doing what we're doing for each major version, hopefully with better automation down the road.

melissa: I was asking if the release package symlinks are going to be just as fragile on yum.puppetlabs.com as you're saying they are on nightlies.puppetlabs.com/yum. I would expect both to be updated when you move to the next version, but from what you're saying, it sounds like yum.puppetlabs.com symlinks will be better maintained?

What is better: grab the release packages from yum.puppetlabs.com or from nightlies.puppetlabs.com/yum?

mwaggett: ah. The fragility I was talking about more general - even though we have no plans to, paths and structure could change, so any assumptions about paths anywhere (nightlies.p.com and yum.pl.com included) that are built into the install function seem fragile to me.

Where you grab release packages from doesn't matter. I was just suggesting pulling from yum since it's already set up in the path you expect.

mwaggett: yum.puppetlabs.com and nightlies.puppet.com/yum should be consistent now - let me know if you notice anything else out of place.

melissa: Thank you!

joshcooper: I'm not sure I understand all of this. Let's check in on Monday.

joshcooper: If I forget to set RELEASE_STREAM then I'm able to download the release package but puppetserver fails to install. Could we make the default `puppet5` so that this works in the default case?

```
Install Puppet Server

  q9be63zfj7ih4mp.delivery.puppetlabs.net (q9be63zfj7ih4mp.delivery.puppetlabs.net) 11:01:10$ rpm --replacepkgs -Uvh http://nightlies.puppet.com/yum/puppet-nightly/puppet-nightly-release-el-7.noarch.rpm
    warning: /var/tmp/rpm-tmp.glHDgk: Header V4 RSA/SHA512 Signature, key ID ef8d349f: NOKEY
    Retrieving http://nightlies.puppet.com/yum/puppet-nightly/puppet-nightly-release-el-7.noarch.rpm
    Preparing...                              ########################################
    Updating / installing...
    puppet-nightly-release-1.0.0-2.el7        ########################################

  q9be63zfj7ih4mp.delivery.puppetlabs.net (q9be63zfj7ih4mp.delivery.puppetlabs.net) executed in 0.30 seconds

  q9be63zfj7ih4mp.delivery.puppetlabs.net (q9be63zfj7ih4mp.delivery.puppetlabs.net) 11:01:10$ echo "/opt/puppetlabs/bin"
    /opt/puppetlabs/bin

  q9be63zfj7ih4mp.delivery.puppetlabs.net (q9be63zfj7ih4mp.delivery.puppetlabs.net) executed in 0.11 seconds

  q9be63zfj7ih4mp.delivery.puppetlabs.net (q9be63zfj7ih4mp.delivery.puppetlabs.net) 11:01:10$ grep ^PATH=.*\/opt\/puppetlabs\/bin ~/.ssh/environment
    PATH=PATH:/opt/puppetlabs/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin

  q9be63zfj7ih4mp.delivery.puppetlabs.net (q9be63zfj7ih4mp.delivery.puppetlabs.net) executed in 0.12 seconds

  q9be63zfj7ih4mp.delivery.puppetlabs.net (q9be63zfj7ih4mp.delivery.puppetlabs.net) 11:01:10$ grep ^PATH=.*PATH ~/.ssh/environment
    PATH=PATH:/opt/puppetlabs/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin

  q9be63zfj7ih4mp.delivery.puppetlabs.net (q9be63zfj7ih4mp.delivery.puppetlabs.net) executed in 0.11 seconds

  q9be63zfj7ih4mp.delivery.puppetlabs.net (q9be63zfj7ih4mp.delivery.puppetlabs.net) 11:01:10$ yum -y  install puppetserver
    Loaded plugins: product-id, search-disabled-repos, subscription-manager
    This system is not registered to Red Hat Subscription Management. You can use subscription-manager to register.
    No package puppetserver available.
    Error: Nothing to do
```

mwaggett: That's because there are no puppetserver packages in the puppet6-nightly repos yet. I'm gonna move that erroneous 6.0.0.master-0.1SNAPSHOT.2018.03.22T1917 package to the right place and then things should work. 

Although I don't think puppetdb has started shipping puppet6 nightlies yet. Not sure if you need those too.

setup/aio/011_Install_Puppet_Server.rb
```diff
@@ -0,0 +1,28 @@
...
+  nightly_builds_url = ENV['NIGHTLY_BUILDS_URL'] || 'http://nightlies.puppet.com'
+  dev_builds_url  = ENV['DEV_BUILDS_URL'] || 'http://builds.delivery.puppetlabs.net'
+
+  if nightly_builds_url == 'http://nightlies.puppet.com'
+    yum_nightlies_url = nightly_builds_url + '/yum'
+    apt_nightlies_url = nightly_builds_url + '/apt'
```
melissa: This isn't quite working for apt

```
irb(main):001:0> require 'uri'
=> true
irb(main):003:0> opts = {}
=> {}
irb(main):004:0> opts[:release_apt_repo_url] = 'http://nightlies.puppet.com/apt'
=> "http://nightlies.puppet.com/apt"
irb(main):005:0> deb = 'puppet5-nightly-release-xenial.deb'
=> "puppet5-nightly-release-xenial.deb"
irb(main):006:0> deb
=> "puppet5-nightly-release-xenial.deb"
irb(main):007:0> opts[:release_apt_repo_url]
=> "http://nightlies.puppet.com/apt"
irb(main):008:0> remote = URI.join( opts[:release_apt_repo_url], deb )
=> #<URI::HTTP http://nightlies.puppet.com/puppet5-nightly-release-xenial.deb>
irb(main):009:0> File.join( opts[:release_apt_repo_url], deb )
=> "http://nightlies.puppet.com/apt/puppet5-nightly-release-xenial.deb"
```

I'll have to update this

melissa: Fixed in https://github.com/puppetlabs/beaker-puppet/pull/30/commits/036eaa106beeb8d11c095b966dfb9b91983ede48

melissa:
I tested this with both a redhat and ubuntu master on 5.3.x, with RELEASE_STREAM='puppet5'. I would still like to test further on the master branch for both those masters and RELEASE_STREAM='puppet6' and not set.

melissa:
I just finished testing both redhat and ubuntu masters on master, with both RELEASE_STREAM='puppet6' and RELEASE_STREAM not set. I feel good about this addition.

joshcooper:
I verified this works correctly with puppet#4.10.x (and SERVER_VERSION=2.8.1), and puppet#5.3.x (with the https://github.com/puppetlabs/puppet/pull/6763 mergeup) both with RELEASE_STREAM not specified (which installed puppetserver 6) and RELEASE_STREAM=puppet5.

PR #31, opened 2018-03-23
(maint) These are changes that were made in later puppet versions tha…

melissa:
…t we should pull back

### DIFF
--- a/setup/git/000_EnvSetup.rb
+++ b/setup/git/000_EnvSetup.rb
@@ -103,10 +103,12 @@ hosts.each do |host|
     arch = host[:ruby_arch] || 'x86'
     step "#{host} Selected architecture #{arch}"
 
+    # TODO we need to be able to branch here. The PA 1.10.x stream needs to use
+    # 2.1.x, and the rest uses 2.4.x
     revision = if arch == 'x64'
-                 '2.1.x-x64'
+                 '2.4.x-x64'
                else
-                 '2.1.x-x86'
+                 '2.4.x-x86'
                end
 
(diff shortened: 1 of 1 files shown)


setup/git/000_EnvSetup.rb
```diff
@@ -103,10 +103,12 @@
...
     step "#{host} Selected architecture #{arch}"
 
+    # TODO we need to be able to branch here. The PA 1.10.x stream needs to use
+    # 2.1.x, and the rest uses 2.4.x
     revision = if arch == 'x64'
-                 '2.1.x-x64'
+                 '2.4.x-x64'
                else
-                 '2.1.x-x86'
+                 '2.4.x-x86'
```
melissa: I pulled this change from https://github.com/puppetlabs/puppet/commit/b548568378841a4497c3238f2f665bc8b4b9617a#diff-77dcb7e76572173a2cfb93ebbf6a5ceb

setup/git/000_EnvSetup.rb
```diff
@@ -132,8 +134,7 @@
 hosts.each do |host|
   case host['platform']
   when /solaris/
-    step "#{host} Install json from rubygems"
-    on host, 'gem install json_pure --no-ri --no-rdoc --version 1.8.3' # json_pure 2.0 requires ruby 2
+    step "#{host} Install bundler from rubygems"
```
melissa: https://github.com/puppetlabs/puppet/commit/a5e7b943104d8e4bf02426d879fefaad62477039

melissa:
These only impact git testing, which is currently broken. This should become unnecessary with the switch to the runtime package.

PR #34, opened 2018-03-30
Maint/master/skip master tests

melissa:


tasks/ci.rake
```diff
@@ -191,7 +196,9 @@ def beaker_suite(type)
   beaker(:init, '--hosts', ENV['HOSTS'], '--options-file', "config/#{String(type)}/options.rb")
   beaker(:provision)
   beaker(:exec, 'pre-suite', '--pre-suite', pre_suites(type))
+  beaker(:exec, 'pre-suite')
```
joshcooper: Why execute `pre-suite` twice?

MikaelSmith: It's applying the pre-suite from beaker-puppet, then any pre-suite in the options file in the component. Mostly useful to pxp-agent, which is doing some extra setup, so that we don't have to have component-specific pre-suite in beaker-puppet.

PR #38, opened 2018-04-11
(BKR-1453) Ensure no agent lock after stopping puppet service

cthorn42:
This PR adds in the agent lock check to after the puppet service
is stopped. There is a window from when we do the initial check on
the agent lock, to where we stop the service.
During that window if a Puppet run goes off the method will end with
the puppet service stopped, but a run in progress. This change will
hopefully cause this to no longer occur.


cthorn42:
@branan is this what you were looking for?

branan:
I don't think there's any point to checking the lock file at all before the service is stopped.

cthorn42:
Good catch.

PR #39, opened 2018-04-11
Maint/master/small changes

melissa:


tasks/ci.rake
```diff
@@ -190,7 +190,9 @@ EOS
       beaker(:provision)
       beaker(:exec, 'pre-suite', '--pre-suite', pre_suites(:gem))
       beaker(:exec, "#{File.dirname(__dir__)}/setup/gem/010_GemInstall.rb")
-      beaker(:destroy) unless ENV['OPTIONS'].include?('--preserve-hosts=always')
+
+      preserve_hosts = ENV['OPTIONS'].include?('--preserve-hosts=always') if ENV['OPTIONS']
```
justinstoller: Should `check_env` do anything with this?

kevpl: @melissa yeah, if `check_env` made sure that `ENV['OPTIONS']` existed, you wouldn't have to check every time you use it. But then again, there are only two uses of it, so it doesn't seem that there are enough to demand that be done there instead of here. lgtm 👍

pcarlisle:
Looks like this PR broke tests

mchllweeks:
@pcarlisle Should be fixed once https://github.com/puppetlabs/beaker-puppet/pull/42 is merged.

PR #40, opened 2018-04-13
Use --module-repository for a custom forge

ekohl:
The current beaker workflow with a custom forge host means stubbing it
in /etc/hosts. This doesn't work because it uses SSL and then the
certificate validation fails. By using the --module-repository option it
just works and even allows you to use plain http and a different port.

Tested with https://github.com/unibet/puppet-forge-server to use caching
to avoid the slow puppet forge.

### DIFF
--- a/lib/beaker-puppet/helpers/puppet_helpers.rb
+++ b/lib/beaker-puppet/helpers/puppet_helpers.rb
@@ -617,6 +617,9 @@ module Beaker
         # forge api v1 canonical source is forge.puppetlabs.com
         # forge api v3 canonical source is forgeapi.puppetlabs.com
         #
+        # @deprecated this method should not be used because stubbing the host
+        # breaks TLS validation.
+        #
         # @param machine [String] the host to perform the stub on
         # @param forge_host [String] The URL to use as the forge alias, will default to using :forge_host in the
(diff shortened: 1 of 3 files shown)


ekohl:
Note there is still some code that stubs the forge URL in /etc/hosts, but this way it's no longer used. That's probably somewhere in beaker itself or another library that parses `BEAKER_FORGE_HOST`.

mchllweeks:
@ekohl lgtm.  Could you please change your commit message to prepend with either a beaker ticket or 'maint'.  Also could you please add a test to ensure we don't break this in the future?

ekohl:
Updated:
* Added `(maint)`
* Rebased
* Added `@deprecated` to forge stubbing functions
* Updated the tests

mchllweeks:

mchllweeks:
Jenkins test this please

mchllweeks:
Jenkins test this please

ekohl:
It'd be good if I could see why the Jenkins test failed.

mchllweeks:
@ekohl The tests are failing due to another issue and not your changes.  Should have these passing again shortly.

mchllweeks:
Jenkins retest this please.

PR #41, opened 2018-04-23
(BKR-1438) Do not use `puppet master`                                …

pcarlisle:
With the removal of webrick the puppet master command no longer works.
This removes calls to `puppet master --configprint`.


tvpartytonight:

PR #43, opened 2018-04-26
Use puppet-agent specific helper when installing the MSI

joshcooper:
This is a partial revert of commit 5a9c90f. When installing
puppet-agent packages we need to use the install_msi_on helper, because
it sets the service startup type to manual, which prevents the service
from starting and sending a CSR before we've configured the agent. It
also verifies that the service startup is correct after installation,
otherwise you can get transient failures if the background agent service
is running while performing puppet agent -t runs in the foreground.

When installing MSI packages for other projects, use the
generic_install_msi_on method.

### DIFF
--- a/lib/beaker-puppet/install_utils/puppet5.rb
+++ b/lib/beaker-puppet/install_utils/puppet5.rb
@@ -103,7 +103,11 @@ module BeakerPuppet
           onhost_package_file = "#{project_name}*"
           host.install_local_package(onhost_package_file)
         when 'windows'
-          generic_install_msi_on(host, artifact_url)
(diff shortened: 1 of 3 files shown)


joshcooper:
I ended up not being able to make larger changes because we need to make sure the master's cert has the required dns alt names.

melissa:
Do you want me to cut a new beaker-puppet release now, or wait?

joshcooper:
Please do @melissa! That should resolve the windows CI failures we're seeing in puppet-agent#5.5.x and master.

melissa:
done!

PR #44, opened 2018-05-01
WIP (maint) Add --allow-insecure-repositories for Ubuntu 18.04 apt commands

ScottGarman:
In Ubuntu 18.04, apt is more strict about using unsigned repos. Since we
don't sign our repos for CI builds, this flag is needed to install our
artifacts during testing.


ScottGarman:
I still need to verify this will actually fix the issue we're encountering, which appears to happen even during apt-get update, and I'm not sure this change impacts how that command is run.

ScottGarman:
Will reopen this with different context.

PR #45, opened 2018-05-01
(CPR-570) Import the gpg signing key on sles 11

melissa:
Please see https://tickets.puppetlabs.com/browse/CPR-570 for additional
details.

Sles 11 has some old and wonky behavior. We can't automatically import
our gpg keys provided with the release package like we do for every
other rpm-based platform. Once we install the release packages, we need
to manually import the keys. This will allow us to install signed
packages from our public repos.

### DIFF
--- a/lib/beaker-puppet/install_utils/foss_utils.rb
+++ b/lib/beaker-puppet/install_utils/foss_utils.rb
@@ -950,6 +950,15 @@ module Beaker
                   [opts[:release_yum_repo_url], repo_name, variant_url_value, version]
               end
 
+              # sles 11 doesn't handle gpg keys well. We can't automatically
+              # import the keys because of sad things, so we have to manually
+              # import it once we install the release package. We'll have to
+              # remember to update this block when we update the signing keys
+              if variant == 'sles' && version == '11'
+                on host, "wget -O /tmp/puppet-gpg-key http://yum.puppetlabs.com/RPM-GPG-KEY-puppet"
+                on host, "rpm --import /tmp/puppet-gpg-key"
+              end
+
               if variant == 'cisco_nexus'
                 # cisco nexus requires using yum to install the repo
                 host.install_package( remote )


lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -962,6 +962,15 @@ def install_puppetlabs_release_repo_on( hosts, repo = nil, opts = options )
...
               end
 
+              # sles 11 doesn't handle gpg keys well. We can't automatically
+              # import the keys because of sad things, so we have to manually
+              # import it once we install the release package. We'll have to
+              # remember to update this block when we update the signing keys
+              if variant == 'sles' && version == '11'
+                on host, "wget -O /tmp/puppet-gpg-key http://yum.puppetlabs.com/RPM-GPG-KEY-puppet"
+                on host, "rpm --import /tmp/puppet-gpg-key"
+              end
```
underscorgan: It looks like this is going to import the gpg key after installing the release package? Might be good to do this before installing the release package, since the release package installation throws up warnings if the key isn't already there, and I found inconsistencies between rpm and zypper for how they worked depending on when you imported the key.

melissa: It doesn't look like it matters? But I'm happy to move this logic


without this change, I get
```
An error occurred in a `before(:suite)` hook.
Failure/Error: run_puppet_install_helper
Beaker::Host::CommandFailure:
  Host 'ajj517057lay7b3.delivery.puppetlabs.net' exited with 4 running:
   zypper --non-interactive --gpg-auto-import-keys in puppet-agent-5.5.1
  Last 10 lines of output were:
  	Abort, retry, ignore? [a/r/i] (a): a
  	Installation of puppet-agent-5.5.1-1.sles11 failed:
  	(with --nodeps --force) Error: Subprocess failed. Error: RPM failed: error: /var/cache/zypp/packages/puppet/puppet-agent-5.5.1-1.sles11.x86_64.rpm: Header V3 RSA/SHA1 signature: BAD, key ID ef8d349f
  	error: /var/cache/zypp/packages/puppet/puppet-agent-5.5.1-1.sles11.x86_64.rpm cannot be installed


  	Problem occured during or after installation or removal of packages:
  	Installation aborted by user

  	Please see the above error message for a hint.
```

With this change, the tests pass as expected

underscorgan: Ok, cool. I may have otherwise confused something in my poking around earlier :)

melissa:
@underscorgan I moved where the key import is happening

lib/beaker-puppet/install_utils/foss_utils.rb
joshcooper: How do we get around this in PE?

PR #48, opened 2018-05-03
(maint) Allow solaris puppet-agent agent-only nodesets

hunner:
This requires a BEAKER_PE_VER anyway.


melissa:
I don't think we should add the ability to install a pe platform from a foss utility. I think I have another solution in the works. I'll open a PR for that shortly against beaker-puppet_install_helper

PR #49, opened 2018-05-03
(BKR-1462) Another Ubuntu 18.04 check to allow insecure repo use

ScottGarman:


speedofdark:
No permissions to review, but (corpsethumb)...

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -1035,6 +1035,10 @@ def install_repo_configs_from_url(host, repo_config_url, copy_dir = nil)
           end
           scp_to( host, repo, to_path )
 
+          if host['platform'] =~ /ubuntu-18.04/
```
MikaelSmith: This appears to make the previous change redundant. The puppet-agent install helper, after setting this, immediately calls code that calls install_repo_configs_from_url.

MikaelSmith: Tying it to apt-get update makes sense. We'll add a repo without signing here, the only other apt-get update corresponds to public repos which should be signed.

MikaelSmith: The most relevant place might be in install_puppetlabs_dev_repo, since that's the specific type of repo we expect to be unsigned.

MikaelSmith:
I'm retesting. @caseywilliams what was the other pipeline that needed the previous fix?

ScottGarman:
Ok, as @MikaelSmith suggested, I've moved the apt config change into install_puppetlabs_dev_repo, and I've also reverted the commit where the first pass attempt was made. Please test this out and let me know if it works.

caseywilliams:
Also still works for puppet-agent acceptance!

ScottGarman:
I've tried to fix the failing PR tests by removing the revert commit and just including the change in this single commit. No difference. Sounds like this is ready for merge now.

MikaelSmith:
Now green.

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -1080,6 +1080,11 @@ def install_puppetlabs_dev_repo ( host, package_name, build_version,
           dev_builds_url = protocol + hostname
           dev_builds_url = opts[:dev_builds_url] if variant =~ /^(fedora|el|centos)$/
 
+          if variant =~ /^ubuntu$/ && version.split('.').first.to_i >= 18
```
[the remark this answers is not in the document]
MikaelSmith: This is now redundant, since we immediately call `install_repo_configs` -> `install_repo_configs_from_url`.

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -1035,6 +1035,11 @@ def install_repo_configs_from_url(host, repo_config_url, copy_dir = nil)
           end
           scp_to( host, repo, to_path )
 
+          variant, version, arch, codename = host['platform'].to_array
+          if variant =~ /^ubuntu$/ && version.split('.').first.to_i >= 18
+            # Allow the use of unsigned repos with Ubuntu 18.04+
+            on host, "echo 'Acquire::AllowInsecureRepositories \"true\";' > /etc/apt/apt     .conf.d/90insecure"
```
caseywilliams: Ah, I should have noticed this before running the tests, but the spaces in the filename are stopping this from working -- this line is applied in the correct tests though.

ScottGarman:
Apologies for that, PR updated.

lib/beaker-puppet/install_utils/foss_utils.rb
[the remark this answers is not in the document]
MikaelSmith: I still really dislike the redundancy here. And I think a spec test covering this would be useful (at the level of: this function ensures that we allow unsigned packages on Ubuntu 18.04).

ScottGarman:
FWIW I've filed a ticket (BKR-1468) to eventually fix this duplication.

PR #62, opened 2018-08-09
(BKR-1509) Beaker 4.0 Compatibility

Dakta:
Assorted fixes for running puppet/acceptance that weren't caught during acceptance testing due to CI configuration conflicts.

### DIFF
--- a/setup/git/010_TestSetup.rb
+++ b/setup/git/010_TestSetup.rb
@@ -1,9 +1,4 @@
 test_name "Install packages and repositories on target machines..." do
-  extend Beaker::DSL::InstallUtils
-
-  SourcePath  = Beaker::DSL::InstallUtils::SourcePath
-  GitHubSig   = Beaker::DSL::InstallUtils::GitHubSig
-
   repositories = options[:install].map do |url|
     extract_repo_info_from(build_git_url(url))
   end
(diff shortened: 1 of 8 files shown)


setup/git/010_TestSetup.rb
```diff
@@ -1,9 +1,4 @@
 test_name "Install packages and repositories on target machines..." do
-  extend Beaker::DSL::InstallUtils
-
-  SourcePath  = Beaker::DSL::InstallUtils::SourcePath
-  GitHubSig   = Beaker::DSL::InstallUtils::GitHubSig
```
melissa: Do we not still need these references defined here?

Dakta: Nope. They're pulled into the DSL namespace, which is set up before pre-suite execution. In fact, these references aren't even complete, since they're relying on a previous chain of inclusion that no longer exists. Both of those constants are actually defined in `FOSSUtils`, which is included in the DSL [here](https://github.com/puppetlabs/beaker-puppet/blob/master/lib/beaker-puppet.rb#L27) via `DSL.register`.

melissa: That's super lovely

Dakta:
Jenkins, retest this please.

PR #66, opened 2018-09-12
(BKR-1528) Use `puppetserver ca` instead of `puppet cert`

Magisus:


lib/beaker-puppet/helpers/puppet_helpers.rb
```diff
@@ -845,8 +858,15 @@ def sign_certificate_for(host = [])
...
+
+              if version_is_less(puppet_version, '5.99')
+                on master, puppet("cert --sign --all --allow-dns-alt-names"), :acceptable_exit_codes => [0,24]
+                out = on(master, puppet("cert --list --all")).stdout
+              else
+                on master, 'puppetserver ca sign --all', :acceptable_exit_codes => [0, 24]
+                out = on(master, 'puppetserver ca list --all').stdout
+              end
+
               if hostnames.all? { |hostname| out =~ /\+ "?#{hostname}"?/ }
```
justinstoller: Did we not need to change this after all?

Magisus: Oh right. Tests are running now. I'll let you know.

Magisus: Kinda looks like this test is just for efficiency. So it may "work" even if the check is failing.

Magisus:
Updated with new regex for 6.0.

PR #71, opened 2018-09-17
Backport additional `puppetserver ca` fixes

Magisus:
This cherry-picks a couple of follow-up commit to the `puppetserver ca` switchover, that are needed for using beaker-puppet in PE with Puppet 6.


steveax:
👍

jpartlow:
What is this branch?

steveax:
This is the branch that older beaker-pe pulls from. Console, Orch, etc. are not on beaker 4 yet and the cert changes are making the install fail at the cert signing step.

jpartlow:
:+1:

PR #72, opened 2018-09-26
(PUP-9136) Ensure state is preserved between runs

melissa:
We need to be able to preserve the state of the hosts that is set up
during the pre-suite runs. We previously were relying on beaker to
initiate this on every invocation, but that behavior was removed in
beaker 4. This commit takes advantage of a new flag that will cause
invocations of `beaker exec` to save the host state. We really only want
to do this when we're running pre-suites.

### DIFF
--- a/beaker-puppet.gemspec
+++ b/beaker-puppet.gemspec
@@ -34,6 +34,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'thin'
 
   # Run time dependencies
+  s.add_runtime_dependency 'beaker', '~> 4.1'
   s.add_runtime_dependency 'stringify-hash', '~> 0.0.0'
   s.add_runtime_dependency 'in-parallel', '~> 0.1'
   s.add_runtime_dependency 'oga'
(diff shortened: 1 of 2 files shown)


melissa:
This depends on https://github.com/puppetlabs/beaker/pull/1544, and may need a beaker bump once that change is merged

melissa:
For those playing along, the PR Tests are currently failing with
```
12:42:59 Bundler could not find compatible versions for gem "beaker":
12:42:59   In Gemfile:
12:42:59     beaker (~> 4.0)
12:42:59 
12:42:59     beaker-puppet was resolved to 1.6.0, which depends on
12:42:59       beaker (>= 4.0.1, ~> 4.0)
12:42:59 
12:42:59 Could not find gem 'beaker (>= 4.0.1, ~> 4.0)', which is required by gem
```

which is perfect, since beaker 4.0.1 should be released tomorrow!! I'll be sure to re-kick these tests once that's available.

melissa:
jenkins please test this

melissa:
jenkins retest this please

PR #76, opened 2018-10-12
(BKR-1523) Add only-fails capability to beaker

speedofdark:
Add rake task that will rerun failures and errors once; depends on
changes in beaker under this same ticket.

### DIFF
--- a/tasks/ci.rake
+++ b/tasks/ci.rake
@@ -3,6 +3,7 @@ require 'pp'
 require 'yaml'
 require 'securerandom'
 require 'fileutils'
+require 'tempfile'
 require 'beaker-hostgenerator'
 require 'beaker/dsl/install_utils'
 extend Beaker::DSL::InstallUtils
@@ -172,14 +173,28 @@ namespace :ci do
 
   namespace :test do
     desc <<-EOS
-Run the acceptance tests using puppet-agent (AIO) packages.
+Run the acceptance tests using puppet-agent (AIO) packages, with or without
+retrying failed tests one time.
 
   $ SHA=<full sha> bundle exec rake ci:test:aio
 
+or
+
+  $ SHA=<full sha> bundle exec rake ci:test:aio[RETRIES]
+
+
 SHA should be the full SHA for the puppet-agent package.
+
+RETRIES should be set to true to retry failed tests one time. It defaults
+to false.
 EOS
-    task :aio => ['ci:check_env', 'ci:gen_hosts'] do
-      beaker_suite(:aio)
+    task :aio, [:retries] => ['ci:check_env', 'ci:gen_hosts'] do |t, args|
+      args.with_defaults(retries: false)
+      if args[:retries]
+        beaker_suite_retry(:aio)
+      else
+        beaker_suite(:aio)
+      end
     end
 
     desc <<-EOS
@@ -193,11 +208,13 @@ EOS
     task :gem => ['ci:check_env'] do
       beaker(:init, '--hosts', 'config/nodes/gem.yaml', '--options-file', 'config/gem/options.rb')
       beaker(:provision)
-      beaker(:exec, 'pre-suite', '--pre-suite', pre_suites(:gem))
-      beaker(:exec, "#{File.dirname(__dir__)}/setup/gem/010_GemInstall.rb")
-
-      preserve_hosts = ENV['OPTIONS'].include?('--preserve-hosts=always') if ENV['OPTIONS']
-      beaker(:destroy) unless preserve_hosts
+      begin
+        beaker(:exec, 'pre-suite', '--preserve-state', '--pre-suite', pre_suites(:gem))
+        beaker(:exec, "#{File.dirname(__dir__)}/setup/gem/010_GemInstall.rb")
+      ensure
+        preserve_hosts = ENV['OPTIONS'].include?('--preserve-hosts=always') if ENV['OPTIONS']
+        beaker(:destroy) unless preserve_hosts
+      end
     end
 
     desc <<-EOS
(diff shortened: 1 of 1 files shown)


tasks/ci.rake
```diff
@@ -193,11 +204,13 @@ EOS
     task :gem => ['ci:check_env'] do
       beaker(:init, '--hosts', 'config/nodes/gem.yaml', '--options-file', 'config/gem/options.rb')
       beaker(:provision)
-      beaker(:exec, 'pre-suite', '--pre-suite', pre_suites(:gem))
-      beaker(:exec, "#{File.dirname(__dir__)}/setup/gem/010_GemInstall.rb")
-
-      preserve_hosts = ENV['OPTIONS'].include?('--preserve-hosts=always') if ENV['OPTIONS']
-      beaker(:destroy) unless preserve_hosts
+      begin
```
speedofdark: this is to make handling gem processing consistent with other c:test options, i.e., always process the destroy regardless of upstream failures.

tasks/ci.rake
```diff
@@ -243,13 +256,37 @@ end
...
+end
+
+def beaker_suite_retry(type)
+  beaker(:init, '--hosts', ENV['HOSTS'], '--options-file', "config/#{String(type)}/options.rb")
+  beaker(:provision)
+  begin
+    beaker(:exec, 'pre-suite', '--preserve-state', '--pre-suite', pre_suites(type))
+    beaker(:exec, 'pre-suite', '--preserve-state')
+
+    beaker(:exec, ENV['TESTS'], '--save-results', '/tmp/results.json')
```
speedofdark: I believe that this should be using something like Tempfile for the interim storage. What do you think? Also, based on my reading of how Tempfile works, if we go this route, I wasn't going to explicitly remove the file and let Tempfile do it's own [cleanup](https://www.rubydoc.info/stdlib/tempfile/Tempfile).

tasks/ci.rake
```diff
@@ -243,13 +256,37 @@ end
...
+  rescue RuntimeError
+    tests_to_rerun = get_tests_from('/tmp/results.json')
```
speedofdark: Another thing I was thinking about was to output the fact that we were retrying, and which spec files were included in the retry?

tasks/ci.rake
```diff
@@ -182,6 +182,17 @@ EOS
...
     end
 
+    desc <<-EOS
+Run the acceptance tests using puppet-agent (AIO) packages with retries.
+
+  $ SHA=<full sha> bundle exec rake ci:test:aioretry
+
+SHA should be the full SHA for the puppet-agent package.
+EOS
+    task :aioretry => ['ci:check_env', 'ci:gen_hosts'] do
```
caseywilliams: I'd call this `:aio_retry` for readability

tasks/ci.rake
caseywilliams: Yes, an auto-created tempfile would make sense. If finding the path for that file again later is an issue, perhaps it could be recorded  in beaker's subcommand_options file on creation?

tasks/ci.rake
```diff
@@ -243,13 +256,37 @@ end
...
+  ensure
+    beaker(:exec, 'post-suite')
+    preserve_hosts = ENV['OPTIONS'].include?('--preserve-hosts=always') if ENV['OPTIONS']
+    beaker(:destroy) unless preserve_hosts
+  end
+end
 
-  preserve_hosts = ENV['OPTIONS'].include?('--preserve-hosts=always') if ENV['OPTIONS']
-  beaker(:destroy) unless preserve_hosts
+def get_tests_from(results_file)
```
caseywilliams: Not sure it's worth adding this method to wrap the JSON load

tasks/ci.rake
```diff
@@ -182,6 +182,17 @@ EOS
...
 
+    desc <<-EOS
+Run the acceptance tests using puppet-agent (AIO) packages with retries.
+
+  $ SHA=<full sha> bundle exec rake ci:test:aioretry
+
+SHA should be the full SHA for the puppet-agent package.
+EOS
+    task :aioretry => ['ci:check_env', 'ci:gen_hosts'] do
+      beaker_suite_retry(:aio)
```
caseywilliams: This is a task that wraps a single method -- would it make sense to just have the retry toggle be a second argument to `beaker_suite`?

kevpl: agreed, seems like doing this would avoid some duplicate code in the methods as well

tasks/ci.rake
speedofdark: Done.

speedofdark:
As to the merge commits, yes, things will be squashed appropriately before hand.

tasks/ci.rake
speedofdark: Refactored.

tasks/ci.rake
speedofdark: Done.

tasks/ci.rake
speedofdark: Done.

tasks/ci.rake
speedofdark: Done

tasks/ci.rake
speedofdark: I think the control flow is sufficiently different that trying to conditionalize (is that a word?) it within one method makes things much less clear than having discrete methods. I agree with don't need a task each; adding a defaulted parameter to :aio would address that.

tasks/ci.rake
speedofdark: Removed aio_retry as a separate task and now use a rake task parameter for :aio instead.

tasks/ci.rake
speedofdark: Defaulted parameter to :aio added.

speedofdark:
retest this please

speedofdark:
retest this please

tasks/ci.rake
```diff
@@ -172,14 +173,28 @@ namespace :ci do
 
   namespace :test do
     desc <<-EOS
-Run the acceptance tests using puppet-agent (AIO) packages.
+Run the acceptance tests using puppet-agent (AIO) packages, with or without
+retrying failed specs one time.
```
ekinanp: Should probably be `failed tests` instead of `failed specs`.

speedofdark:
The above failure is not related to PR based changes.

tasks/ci.rake
```diff
@@ -243,9 +260,10 @@ end
 def beaker_suite(type)
```
ekinanp: Can we just make the retry optional here? E.g.
```
def beaker_suite(type, options = {})
end
```

then you can do
```
rescue RuntimeError => e
  raise e unless options[:retry]
  // retry code
end
```

speedofdark: Thanks! I'll look at that option for the next release. In theory, we might be able to ship this tomorrow; makes me reluctant to have to change things right now.

PR #77, opened 2018-10-18
(PA-2183) Updates to get agent CI  working with Beaker 4

ekinanp:


lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -1009,6 +1009,8 @@ def install_puppetlabs_release_repo_on( hosts, repo = nil, opts = options )
             end
             configure_type_defaults_on( host )
```
ekinanp: Would it be better to put the `write_hosts` call in `configure_type_defaults_on` ? I can see arguments for and against this. Until that is settled, I have chosen not to update the unit tests with `write_hosts` in this PR (hence the default stubbing).

lib/beaker-puppet/helpers/facter_helpers.rb
```diff
@@ -44,7 +44,7 @@ def fact_on(host, name, opts = {})
             opts << ' --json'
           end
 
-          result = on host, facter(name, opts)
+          result = on host, facter("\"#{name}\"", opts)
```
ekinanp: An example of a fact name with spaces is https://github.com/ekinanp/facter/commit/31d34005c14d9c33ed1ab42b0d1810f5b9412df0#diff-1796fc5451255bbafbb615c5ce6ed0faR29 where on Windows, `primary_interface` has a space in it.

ekinanp:
This is also blocked on https://github.com/puppetlabs/beaker/pull/1551

lib/beaker-puppet/install_utils/foss_utils.rb
ekinanp: Though it would work if write_hosts were modified to only update the given hosts in the `HOSTS` entry. I'll probably do that once my ad-hoc run passes.

ekinanp: Updated the PR to do this instead.

PR #79, opened 2018-10-23
(PA-2183) Quote fact name in fact_on helper

ekinanp:
The fact name can have spaces. Thus, it should be quoted before it is passed
into the shell.


ekinanp:
An example of a fact name with spaces is ekinanp/facter@31d3400#diff-1796fc5451255bbafbb615c5ce6ed0faR29 where on Windows, primary_interface has a space in it.

PR #82, opened 2018-11-19
(maint) If iptables doesn't exist, don't fail

melissa:
If we're on a system that should have iptables but doesn't for some
reason, then we don't need to purge iptables.

This happens on systems like docker, where some images just don't have
that. We don't need to manage it.


melissa:
jenkins retest this please

PR #83, opened 2018-11-23
Autodetect the target module dir in copy_module_to

ekohl:
When using `BEAKER_provision=no` the `host['distmoduledir']` is undefined. This means we need to autodetect it. For this we can let puppet print its config for us. This is in fact a much more fool proof method than relying on some variable so we prefer this now.

Also removes an unused variable.

### DIFF
--- a/lib/beaker-puppet/install_utils/module_utils.rb
+++ b/lib/beaker-puppet/install_utils/module_utils.rb
@@ -120,7 +120,6 @@ module Beaker
             ignore_list = build_ignore_list(opts)
             target_module_dir = on( host, "echo #{opts[:target_module_path]}" ).stdout.chomp
             source_path = File.expand_path( opts[:source] )
-            source_dir = File.dirname(source_path)
             source_name = File.basename(source_path)
             if opts.has_key?(:module_name)
               module_name = opts[:module_name]
--- a/lib/beaker-puppet/install_utils/module_utils.rb
+++ b/lib/beaker-puppet/install_utils/module_utils.rb
@@ -118,7 +118,7 @@ module Beaker
                     :ignore_list => PUPPET_MODULE_INSTALL_IGNORE}.merge(opts)
 
             ignore_list = build_ignore_list(opts)
-            target_module_dir = on( host, "echo #{opts[:target_module_path]}" ).stdout.chomp
+            target_module_dir = get_target_module_path(host, opts[:target_module_path])
             source_path = File.expand_path( opts[:source] )
             source_name = File.basename(source_path)
             if opts.has_key?(:module_name)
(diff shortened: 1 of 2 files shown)


sbeaulie:
ok to test

lib/beaker-puppet/install_utils/module_utils.rb
```diff
@@ -114,13 +114,16 @@ def install_puppet_module_via_pmt( opts = {} )
         def copy_module_to(one_or_more_hosts, opts = {})
           block_on one_or_more_hosts do |host|
             opts = {:source => './',
-                    :target_module_path => host['distmoduledir'],
                     :ignore_list => PUPPET_MODULE_INSTALL_IGNORE}.merge(opts)
 
+            target_module_dir = if opts[:target_module_path]
+                                  on( host, "echo #{opts[:target_module_path]}" ).stdout.chomp
+                                else
+                                  on( host, puppet("config print basemodulepath") ).stdout.chomp.split(':').first
```
sbeaulie: I have seen `puppet config` print some warning that could impact this in non-happy path scenarios:
```
$ puppet config print basemodulepath
WARN: Clearing out unresolved specs.
Please report a bug if this causes problems.
/Users/$foo/modules:/opt/puppetlabs/puppet/modules
```
Which would end up copying to "WARN/$module_name"
It may be worth checking a valid path is captured before continuing with the copy. Or only process the last line of stdout?

lib/beaker-puppet/install_utils/module_utils.rb
```diff
@@ -114,13 +114,16 @@ def install_puppet_module_via_pmt( opts = {} )
         def copy_module_to(one_or_more_hosts, opts = {})
           block_on one_or_more_hosts do |host|
             opts = {:source => './',
-                    :target_module_path => host['distmoduledir'],
```
sbeaulie: In this version of code we had this default, which is lost (to the autodetect)

ekohl: I considered autodetect to be better, but I don't feel strongly about this.

lib/beaker-puppet/install_utils/module_utils.rb
ekohl: If Puppet is well behaved then it'll output warnings on stderr and it should be no problem but it's a good point. It could also show a deprecation warning. Checking if it's a valid path is a good suggestion and I'll look into this.

lib/beaker-puppet/install_utils/module_utils.rb
ekohl: I have a commit locally that reverts to the old behavior but it also includes the path checking code which I haven't had time to test yet.

ekohl:
Updated. Now with tests.

ekohl:
@kevpl mind having another look?

lib/beaker-puppet/install_utils/module_utils.rb
ekohl: Checking if the directory exists doesn't really work because this code is executed on the host, not on the client. There are also cases where a directory in the modulepath might not exist yet but can be created. I'd say that variable interpolation is not supported and needs to be passed in explicitly.

ekohl:
I've updated it to use the `host.puppet` helper to remove some duplicate code and dropped the file existence checks because they don't work.

kevpl:
this looks fine to me. Please pre-pend your commit messages with `(MAINT)` or a BKR ticket ref, & we'll get this pushed through to merge & release!

ekohl:
Updated with a new commit message

kevpl:
the failure is known & unrelated to your changes, so I think we're good here. Releasing after merge

PR #88, opened 2018-11-29
(BKR-1558) make sure we destroy hosts for ci:test:* task execution

speedofdark:
This wraps the destroy subcommand in an ensure block to make sure
that we always get rid of hosts even if other subcommands cause a
rake failure after provisioning.

### DIFF
--- a/tasks/ci.rake
+++ b/tasks/ci.rake
@@ -243,13 +243,15 @@ end
 def beaker_suite(type)
   beaker(:init, '--hosts', ENV['HOSTS'], '--options-file', "config/#{String(type)}/options.rb")
   beaker(:provision)
-  beaker(:exec, 'pre-suite', '--preserve-state', '--pre-suite', pre_suites(type))
-  beaker(:exec, 'pre-suite', '--preserve-state')
-  beaker(:exec, ENV['TESTS'])
-  beaker(:exec, 'post-suite')
-
-  preserve_hosts = ENV['OPTIONS'].include?('--preserve-hosts=always') if ENV['OPTIONS']
-  beaker(:destroy) unless preserve_hosts
+  begin
+    beaker(:exec, 'pre-suite', '--pre-suite', pre_suites(type))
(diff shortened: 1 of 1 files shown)


sbeaulie:
retest this please

speedofdark:
As this will markedly change behavior in terms of keeping hosts around when tests fail, do we need to communicate about using --preserve-hosts more widely?

mchllweeks:
@speedofdark We are trying to understand the change you are requesting and the repercussions it might have better.  Is this for local testing or for CI?  Could you add more detail why this change is necessary?

speedofdark:
These changes are targeted for a CI context; when running any of the ci:test:* test tasks (except gem), the sequence of beaker subcommands to run is defined by [beaker_suite](https://github.com/puppetlabs/beaker-puppet/blob/49bf905869d564f1c6c157f661a92e0f709d2925/tasks/ci.rake#L243-L253), which in turn uses [beaker](https://github.com/puppetlabs/beaker-puppet/blob/49bf905869d564f1c6c157f661a92e0f709d2925/tasks/ci.rake#L237-L241) and then [sh](https://github.com/puppetlabs/beaker-puppet/blob/49bf905869d564f1c6c157f661a92e0f709d2925/tasks/ci.rake#L243). The sh command will exit with a status of 1 when tests fail. This in turn aborts rake, when then prevents any of the commands following test execution to be run.

In this specific instance, when tests fail, we don't run post-suite or de-provision machines via [destroy](https://github.com/puppetlabs/beaker-puppet/blob/49bf905869d564f1c6c157f661a92e0f709d2925/tasks/ci.rake#L252). The end result is that when tests fail, we leave machines provisioned until they are reaped.

speedofdark:
I'll open a different PR to update the command sequencing for gems.

speedofdark:
This page didn't get the updated commit message, but if you go over to commits and clock through, you'll see the updated message.

speedofdark:
retest this please

PR #90, opened 2019-01-03
(BKR-1560) Move server setup and sign cert tasks to their own helpers

caseywilliams:
In the puppet-agent acceptance suites, puppetserver is installed and agent certs are signed using pre-suite tests from beaker-puppet.

These changes extract the logic from those pre-suite tests into reusable helpers in the DSL:

- `setup/common/011_Install_Puppet_Server.rb` now calls `install_puppetserver_on`
- `setup/common/040_ValidateSignCert.rb` now calls `sign_agent_certs`

### DIFF
--- a/spec/beaker-puppet/install_utils/foss_utils_spec.rb
+++ b/spec/beaker-puppet/install_utils/foss_utils_spec.rb
@@ -1444,6 +1444,93 @@ describe ClassMixedWithDSLInstallUtils do
     end
   end
 
+  describe '#install_puppetserver_on' do
+    let(:host) { make_host('master', platform: 'el-7-x86_64') }
+
+    context 'with default arguments' do
+      it "installs puppetserver 2.x from the default 'puppet' release stream" do
+        expect(subject).to receive(:install_puppetlabs_release_repo_on).with(host, 'puppet')
+        expect(subject).to receive(:install_package).with(host, 'puppetserver')
+        subject.install_puppetserver_on(host)
+      end
+    end
+
+    context 'with a version option' do
+      version = '6.6.6'
+
+      it 'installs puppetserver at the specific version from internal buildservers' do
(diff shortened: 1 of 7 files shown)


spec/beaker-puppet/install_utils/foss_utils_spec.rb
```diff
@@ -1444,6 +1444,93 @@ def test_fetch_http_file_no_ending_slash(platform)
...
+
+        it 'uses a custom general nightlies url' do
+          expect(subject).to receive(:install_puppetlabs_release_repo_on)
+                                 .with(host, 'puppet-nightly',
+                                       release_yum_repo_url: nightlies_url,
+                                       release_apt_repo_url: nightlies_url)
+          subject.install_puppetserver_on(host, nightlies: true, nightly_builds_url: nightlies_url)
+        end
+
+        it 'uses a custom yum repo url' do
```
sbeaulie: just a typo it seems the test description should be flipped with the next one apt/yum

kevpl:
also, the testing failure is a known & unrelated issue, so that shouldn't block this moving forward, FYI

caseywilliams:
Yay, thanks for looking @kevpl and @sbeaulie -- I'll fix the typo, and I think I'll get a review or two from my team before I remove the `do not merge`.

Side note: I tested these changes with the puppet-agent pipelines for puppet 1.10.x, 5.5.x, and 6.0.x -- everything still seems to work fine there.

kevpl:
apologies, didn't see the label. Will wait for the changes then.

kevpl:
@caseywilliams how are we looking here? It looks like the last change was 6 days ago, but 3 days ago you proposed other work that depends on this? Are we ready for the `do not merge` label to be taken off & to get another look?

caseywilliams:
We are not at a good spot to use this work in the puppet_agent module or the puppet-agent pipelines, and I just don't have time to finish this right now. I'm going to close this PR and open a less invasive one that will cover our needs for now.

PR #92, opened 2019-01-17
(maint) Fix empty argument case in collection methods

caseywilliams:
Stop puppet collection methods from erroring without an argument


caseywilliams:
PR test failure seems unrelated:
```
07:51:54 + env bundle exec rspec -fd -c spec
07:51:54 bundler: command not found: rspec
07:51:54 Install missing gem executables with `bundle install`
07:51:54 Build step 'Execute shell' marked build as failure
07:51:55 Unable to get pull request builder trigger!!
```

PR #95, opened 2019-01-17
(BKR-1560) Add install_puppetserver_on helper

caseywilliams:
puppet-agent's test suite relies on a reusable test file,
setup/common/011_Install_Puppet_Server.rb, to install puppetserver on
puppet masters before the tests run. This extracts the logic from that
test file into a helper method, `install_puppetserver_on`, which does
the same thing.

### DIFF
--- a/spec/beaker-puppet/install_utils/foss_utils_spec.rb
+++ b/spec/beaker-puppet/install_utils/foss_utils_spec.rb
@@ -1444,6 +1444,93 @@ describe ClassMixedWithDSLInstallUtils do
     end
   end
 
+  describe '#install_puppetserver_on' do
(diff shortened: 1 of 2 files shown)


spec/beaker-puppet/install_utils/foss_utils_spec.rb
```diff
@@ -1444,6 +1444,93 @@ def test_fetch_http_file_no_ending_slash(platform)
     end
   end
 
+  describe '#install_puppetserver_on' do
+    let(:host) { make_host('master', platform: 'el-7-x86_64') }
+
+    context 'with default arguments' do
+      it "installs puppetserver 2.x from the default 'puppet' release stream" do
```
caseywilliams: oops, this description is wrong -- updating.

PR #96, opened 2019-01-22
(maint) Fix incorrect constant reference, update a spec test description

caseywilliams:
Not sure how the constant reference passed PR spec tests in the first place -- I've also included an update to a test description that was incorrect, although the behavior was fine.


speedofdark:
PR tests are dying because 18n doesn't like the ruby version?

caseywilliams:
oh right. it's because the PR tests are broken

PR #97, opened 2019-01-23
A few small puppet helper fixes

caseywilliams:
Three small commits here:

- Use the `facter` helper in `puppet_agent_version_on` - This seems to be more reliably correct when executing facter on windows where the PATH may not be updated correctly on the SUT.
- Consistently downcase collection names - There are no puppet collections that use capital letters, so always convert any collection parameters to `install_puppet_agent_on` to lowercase (instead of expecting the caller to get it right -- this is often confusing for PC1/pc1)
- Pass host targets along to the install method during puppetserver install (my mistake)


caseywilliams:
PR tests are still broken, but spec test pass for me locally

PR #98, opened 2019-01-23
(maint) Enable acceptance testing with puppet-agent installed from ni…

melissa:
…ghtlies

### DIFF
--- a/setup/aio/010_Install_Puppet_Agent.rb
+++ b/setup/aio/010_Install_Puppet_Agent.rb
@@ -1,10 +1,23 @@
 test_name "Install Packages" do
 
   dev_builds_url  = ENV['DEV_BUILDS_URL'] || 'http://builds.delivery.puppetlabs.net'
+  nightly_builds_url = ENV['NIGHTLY_BUILDS_URL'] || 'http://nightlies.puppet.com'
+
+  sha = ENV['SHA']
+  release_stream = ENV['RELEASE_STREAM'] || 'puppet'
 
   step "Install puppet-agent..." do
-    sha = ENV['SHA']
-    install_from_build_data_url('puppet-agent', "#{dev_builds_url}/puppet-agent/#{sha}/artifacts/#{sha}.yaml", hosts)
+    # If SHA='latest', then we're installing from nightlies
+    if sha == 'latest'
+      opts = {
+        :release_yum_repo_url => nightly_builds_url + '/yum',
+        :release_apt_repo_url => nightly_builds_url + '/apt',
+        :puppet_collection => "#{release_stream}-nightly"
+      }
+      install_puppet_agent_on(hosts, opts)
+    else
+      install_from_build_data_url('puppet-agent', "#{dev_builds_url}/puppet-agent/#{sha}/artifacts/#{sha}.yaml", hosts)
+    end
   end
 
   # make sure install is sane, beaker has already added puppet and ruby
--- a/lib/beaker-puppet/install_utils/foss_utils.rb
+++ b/lib/beaker-puppet/install_utils/foss_utils.rb
@@ -400,7 +400,7 @@ module Beaker
             end
 
             if package_name
-              install_puppetlabs_release_repo( host, opts[:puppet_collection] )
+              install_puppetlabs_release_repo( host, opts[:puppet_collection] , opts)
               host.install_package( package_name )
             end
           end


setup/aio/010_Install_Puppet_Agent.rb
```diff
@@ -1,10 +1,23 @@
...
+  nightly_builds_url = ENV['NIGHTLY_BUILDS_URL'] || 'http://nightlies.puppet.com'
+
+  sha = ENV['SHA']
+  release_stream = ENV['RELEASE_STREAM'] || 'puppet'
 
   step "Install puppet-agent..." do
-    sha = ENV['SHA']
-    install_from_build_data_url('puppet-agent', "#{dev_builds_url}/puppet-agent/#{sha}/artifacts/#{sha}.yaml", hosts)
+    # If SHA='latest', then we're installing from nightlies
+    if sha == 'latest'
```
melissa: actually, I want to check both SHA and RELEASE_STREAM. If RELEASE_STREAM is set to something like, `puppet5-nightly` (i.e., ends with -nightly), then that should also trigger this block

melissa: Nope, I changed my mind. That adds too much complication, especially for this round. I think this is fine as is.

melissa:
I tested this successfully with redhat7, but windows, ubuntu, and debian have both failed because of infrastructure/availability issues

melissa:
Ruby 2.2.5 failed with 
```
13:31:36 An error occurred while installing i18n (1.5.3), and Bundler cannot continue.
13:31:36 Make sure that `gem install i18n -v '1.5.3' --source
13:31:36 'https://artifactory.delivery.puppetlabs.net/artifactory/api/gems/rubygems/'`
13:31:36 succeeds before bundling.
13:31:36 
13:31:36 In Gemfile:
13:31:36   markdown was resolved to 1.2.0, which depends on
13:31:36     textutils was resolved to 1.4.0, which depends on
13:31:36       activesupport was resolved to 5.2.2, which depends on
13:31:36         i18n
```

melissa:
Jenkins please retest this

melissa:
Acceptance tests have been broken for a while, and can be ignored

setup/aio/010_Install_Puppet_Agent.rb
```diff
@@ -1,10 +1,23 @@
...
+      opts = {
+        :release_yum_repo_url => nightly_builds_url + '/yum',
+        :release_apt_repo_url => nightly_builds_url + '/apt',
+        :puppet_collection => "#{release_stream}-nightly"
+      }
+      install_puppet_agent_on(hosts, opts)
+    else
+      install_from_build_data_url('puppet-agent', "#{dev_builds_url}/puppet-agent/#{sha}/artifacts/#{sha}.yaml", hosts)
```
caseywilliams: I added a helper to wrap this recently (so that the artifactory doesn't need to be specified) if that helps: https://github.com/puppetlabs/beaker-puppet/commit/fef05c36fe53fecc01482ba08c93ee8217682bed

melissa: This is why I should have been paying more attention to the changes y'al made to this code base! That's p sweet

melissa: @caseywilliams I'm curious about this helper you wrote

beaker already has a utility to check the accessibility of a url that gets called in `install_from_build_data_url`. Is there a reason you added a brand new check with `dev_builds_accessible_on?`? The main thing I can see is if you're running tests on a host that doesn't have access to builds.d.pl.net versus the local system that's running beaker.

The other thing is that you can't pass in a custom url. I know it's super unlikely, but there is a possibility that folks have duplicated how we've set up builds.d.pl.net with the yaml file. Since we're already giving people a means to customize that url, I'm hesitant to take that functionality away

caseywilliams: I intended for the dev builds accessibility helper to be a good-enough check that would determine whether the hosts can access resources within the corporate VPN, regardless of whether you were actually trying to download something at the time (like for printing human-friendly error messages), but now that you put it that way, I realize the name is bad and it's probably not even that useful :/

This PR does look good to me

melissa: It is helpful, and it does exactly what you would expect based on the name!

Mostly I noticed beaker has the method `link_exists?` at https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/helpers/web_helpers.rb#L20. I was curious why you weren't using that

caseywilliams: Ah yeah, I got some feedback on the original PR that encouraged checking all of the SUTs' access as opposed to just the controlling host

kevpl:
seems good to me. @melissa any reason we shouldn't merge & release this now?

caseywilliams:
I've tested the normal (not nightly) package installation with ad-hoc agent acceptance, and everything still seems to work fine there, fwiw

melissa:
@caseywilliams are we in the clear with releases to merge this in?

caseywilliams:
Yes, now is a good time to merge this!

PR #100, opened 2019-01-24
(maint) Centralize puppet_collection_for logic

ekinanp:
This commit centralizes the puppet_collection_for_puppet_agent_version
and puppet_collection_for_puppet_version logic into a single
puppet_collection_for helper.

### DIFF
--- a/lib/beaker-puppet/install_utils/puppet_utils.rb
+++ b/lib/beaker-puppet/install_utils/puppet_utils.rb
@@ -90,41 +90,51 @@ module Beaker
           collection
         end
 
-        # Determine the puppet collection that matches a given version of the puppet-agent
-        # package (you can find this version in the `aio_agent_version` fact).
+        # Determine the puppet collection that matches a given package version. The package
+        # must be one of
+        #   * :puppet_agent (you can get this version from the `aio_agent_version_fact`)
+        #   * :puppet
         #
-        # @param agent_version [String] a semver version number of the puppet-agent package, or the string 'latest'
-        # @returns [String|nil] the name of the corresponding puppet collection, if any
-        def puppet_collection_for_puppet_agent_version(agent_version)
-          agent_version = agent_version.to_s
-          return 'puppet' if agent_version.strip == 'latest'
-
-          x, y, z = agent_version.to_s.split('.').map(&:to_i)
-          return nil if x.nil? || y.nil? || z.nil?
-
-          return 'pc1' if x == 1
-
-          # A y version >= 99 indicates a pre-release version of the next x release
-          x += 1 if y >= 99
-          "puppet#{x}" if x > 4
-        end
-
-        # Determine the puppet collection that matches a given version of the puppet gem.
         #
-        # @param version [String] a semver version number of the puppet gem, or the string 'latest'
+        # @param package [Symbol] the package name. must be one of :puppet_agent, :puppet
+        # @param version [String] a version number or the string 'latest'
         # @returns [String|nil] the name of the corresponding puppet collection, if any
-        def puppet_collection_for_puppet_version(puppet_version)
-          puppet_version = puppet_version.to_s
-          return 'puppet' if puppet_version.strip == 'latest'
+        def puppet_collection_for(package, version)
+          valid_packages = [
+            :puppet_agent,
+            :puppet
+          ]
 
-          x, y, z = puppet_version.to_s.split('.').map(&:to_i)
-          return nil if x.nil? || y.nil? || z.nil?
+          unless valid_packages.include?(package)
+            raise "package must be one of #{valid_packages.join(', ')}"
+          end
 
-          return 'pc1' if x == 4
+          case package
+          when :puppet_agent
+            agent_version = version.to_s
+            return 'puppet' if agent_version.strip == 'latest'
 
-          # A y version >= 99 indicates a pre-release version of the next x release
-          x += 1 if y >= 99
-          "puppet#{x}" if x > 4
+            x, y, z = agent_version.to_s.split('.').map(&:to_i)
+            return nil if x.nil? || y.nil? || z.nil?
+
+            return 'pc1' if x == 1
+
+            # A y version >= 99 indicates a pre-release version of the next x release
+            x += 1 if y >= 99
+            "puppet#{x}" if x > 4
+          when :puppet
+            puppet_version = version.to_s
+            return 'puppet' if puppet_version.strip == 'latest'
+
+            x, y, z = puppet_version.to_s.split('.').map(&:to_i)
+            return nil if x.nil? || y.nil? || z.nil?
+
+            return 'pc1' if x == 4
+
+            # A y version >= 99 indicates a pre-release version of the next x release
+            x += 1 if y >= 99
+            "puppet#{x}" if x > 4
+          end
         end
 
         # Report the version of puppet-agent installed on `host`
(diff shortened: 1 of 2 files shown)


lib/beaker-puppet/install_utils/puppet_utils.rb
```diff
@@ -90,41 +90,51 @@ def get_puppet_collection(agent_version = 'latest')
...
         #
-        # @param agent_version [String] a semver version number of the puppet-agent package, or the string 'latest'
+        #
+        # @param package [Symbol] the package name. must be one of :puppet_agent, :puppet
+        # @param version [String] a version number or the string 'latest'
         # @returns [String|nil] the name of the corresponding puppet collection, if any
-        def puppet_collection_for_puppet_agent_version(agent_version)
-          agent_version = agent_version.to_s
-          return 'puppet' if agent_version.strip == 'latest'
+        def puppet_collection_for(package, version)
```
ekinanp: Chose this instead of `puppet_collection_for(agent_version: <version>)` because a single entry hash parameter doesn't really make sense.

lib/beaker-puppet/install_utils/puppet_utils.rb
```diff
@@ -90,41 +90,51 @@ def get_puppet_collection(agent_version = 'latest')
...
+          valid_packages = [
+            :puppet_agent,
+            :puppet
```
melissa: Shouldn't we have a longer list of valid packages here? Puppetserver, puppetdb, puppetdb-termini are all versioned on the same scheme as puppet and puppet_agent, so there is a case to be made that we might need to find the release stream for those packages.

http://apt.puppet.com/pool/jessie/puppet6/p/

lib/beaker-puppet/install_utils/puppet_utils.rb
```diff
@@ -90,41 +90,51 @@ def get_puppet_collection(agent_version = 'latest')
...
+          when :puppet_agent
+            agent_version = version.to_s
+            return 'puppet' if agent_version.strip == 'latest'
 
-          # A y version >= 99 indicates a pre-release version of the next x release
-          x += 1 if y >= 99
-          "puppet#{x}" if x > 4
-        end
+            x, y, z = agent_version.to_s.split('.').map(&:to_i)
+            return nil if x.nil? || y.nil? || z.nil?
```
melissa: Is there a way we can reduce the redundancies here? The only real difference between `:puppet_agent` and `:puppet` that I can tell is the pc1 logic (`return 'pc1' if x == 1` vs `return 'pc1' if x == 4`).

lib/beaker-puppet/install_utils/puppet_utils.rb
```diff
@@ -90,41 +90,51 @@ def get_puppet_collection(agent_version = 'latest')
...
+            x += 1 if y >= 99
+            "puppet#{x}" if x > 4
+          when :puppet
+            puppet_version = version.to_s
+            return 'puppet' if puppet_version.strip == 'latest'
 
-          x, y, z = puppet_version.to_s.split('.').map(&:to_i)
-          return nil if x.nil? || y.nil? || z.nil?
+            x, y, z = puppet_version.to_s.split('.').map(&:to_i)
+            return nil if x.nil? || y.nil? || z.nil?
```
melissa: Hmm, I feel like we have enough information to return a valid result if the only thing we have is `x`, since it's only the major version that defines the release stream. Though, I suppose if `y` and `z` are nil that might indicate a malformed version string.... I'm not sure my hang up on this is worth any follow up work

lib/beaker-puppet/install_utils/puppet_utils.rb
```diff
@@ -90,41 +90,51 @@ def get_puppet_collection(agent_version = 'latest')
...
 
-          return 'pc1' if x == 4
+            return 'pc1' if x == 4
```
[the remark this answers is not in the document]
melissa: Is beaker-puppet used int he puppet_agent module? That's a little scary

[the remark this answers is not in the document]
melissa: oh wait, I shouldn't be surprised by this. The beaker puppet install helper uses beaker-puppet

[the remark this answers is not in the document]
melissa: However, wouldn't this only be needed if we were installing new package? This should be verified, but I'm not convinced we need this for packages that already exist on the system we are upgrading. Just a thought. It would be nice to start removing all the `pc1` references in this code. It's totes something we can save for later.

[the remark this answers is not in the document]
ekinanp: We do need PC1 in the `:puppet_agent` package case because the puppet_agent module tests calculate the collection from the agent version. I don't know about PC1 in the `:puppet` package case though. @caseywilliams ?

lib/beaker-puppet/install_utils/puppet_utils.rb
ekinanp: I was thinking this could be something that'd be extended on a per-need case? So if a puppetserver person finds out that they need to calculate the collection from their package version, then they can extend this method with that code (and they'd figure out they need the collection from the raised exception). If you think it is a good idea to do as much as we can in the current implementation, however, I'm happy to update it. I'm not sure what the collection should be for legacy server/puppetdb versions (like Server 2.x, since agent 1.x uses it). Maybe we don't need to worry about those.

lib/beaker-puppet/install_utils/puppet_utils.rb
ekinanp: Yeah, I can go ahead and do that.

ekinanp: Updated the PR.

lib/beaker-puppet/install_utils/puppet_utils.rb
ekinanp: We'll still need `y` to handle weird cases like the `5.99` tag (I forgot why we did that in the first place -- maybe @caseywilliams can clarify ?).

It might be worth raising an exception for a malformed version string?

lib/beaker-puppet/install_utils/puppet_utils.rb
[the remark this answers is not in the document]
melissa: 👍 sounds good :) Safest to keep it

lib/beaker-puppet/install_utils/puppet_utils.rb
melissa: I suppose it depends on where this method is used? But generally since this repo is so ad hoc, I'm fine waiting to add those other packages until we need them. As long as that's a conscious decision

lib/beaker-puppet/install_utils/puppet_utils.rb
melissa: I feel like raising an error isn't a bad idea if we can't figure out what's happening with the version string. Do we have existing utilities available to help parse version strings?

ekinanp: There is [Beaker::Shared::Semvar](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/shared/semvar.rb#L12), but that doesn't raise any errors if the versions are malformed strings and it isn't specific to semver parsing.

ekinanp:
Sorry, I realized there were some dangling references to the old puppet_collection_for_<package> methods. Pushed up a commit to remove them.

melissa:
I manually edited the PR Tests job for this PR so I could test https://github.com/puppetlabs/ci-job-configs/pull/5530, but at least we know that's passing!

PR #102, opened 2019-01-26
(maint) Avoid mutating frozen string

donoghuc:
Avoid risk of RunTime errors related to mutating frozen strings by not modifying puppet_collection in place.

### DIFF
--- a/lib/beaker-puppet/install_utils/foss_utils.rb
+++ b/lib/beaker-puppet/install_utils/foss_utils.rb
@@ -362,7 +362,8 @@ module Beaker
           opts[:puppet_agent_version] ||= opts[:version] #backwards compatability with old parameter name
           opts[:puppet_collection] ||= puppet_collection_for(:puppet_agent, opts[:puppet_agent_version]) || 'pc1'
 
-          opts[:puppet_collection].downcase! # the collection names are case sensitive
+          # the collection names are case sensitive
+          opts[:puppet_collection] = opts[:puppet_collection].downcase
 
           run_in_parallel = run_in_parallel? opts, @options, 'install'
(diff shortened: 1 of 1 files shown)


donoghuc:
Noticed this with version `1.15.0`. Example: https://jenkins-master-prod-1.delivery.puppetlabs.net/job/forge-module_puppetlabs-service_intn-sys_smoke-adhoc/PLATFORM=ubuntu1604-64default.a,WORKER_LABEL=beaker/18/console

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -362,7 +362,8 @@ def install_puppet_agent_on(hosts, opts = {})
           opts[:puppet_agent_version] ||= opts[:version] #backwards compatability with old parameter name
           opts[:puppet_collection] ||= puppet_collection_for(:puppet_agent, opts[:puppet_agent_version]) || 'pc1'
```
ekohl: Should `puppet_collection_for` return a downcased collection instead?

bastelfreak:
Hi people. I tested this fix and can confirm that it works for the Vox Pupuli org. I also suggest adding `# frozen_string_literal: true` to the file header.

lib/beaker-puppet/install_utils/foss_utils.rb
kevpl: fair question, but since there is some time pressure due to caused failures from the earlier change & what you're suggesting is functionally equivalent, I'm going to move forward with merging & releasing this change as-is.

Please propose a new PR with your suggested change if you'd like to pursue this further in a later release.

PR #103, opened 2019-01-28
(maint) `agent` should be `host`

melissa:
It turns out `agent` and `agents` are equivalent if you don't explicitly set
`agent` to something. Who knew! The loop that this exists in we refer to
the individual servers as `host`, not `agent` as I had previously, and
incorrectly, assumed.

This resulted in some strange errors depending on how many agents we had
defined for a given run. `agent`, and I assume `agents` as well, returns
a host object if there is only one host with the agent role and it will
return an array of hosts if there are multiple hosts with the agent
role.

### DIFF
--- a/setup/git/010_TestSetup.rb
+++ b/setup/git/010_TestSetup.rb
@@ -22,8 +22,8 @@ test_name "Install repositories on target machines..." do
       #           host_path: ~/puppet
       #           container_path: /build/puppet
       #
-      if agent[:mount_folders]
-        mount = agent[:mount_folders][repository[:name]]
+      if host[:mount_folders]
+        mount = host[:mount_folders][repository[:name]]
         repository[:path] = "file://#{mount[:container_path]}"
       end
       repo_dir = host.tmpdir(repository[:name])


setup/git/010_TestSetup.rb
```diff
@@ -22,7 +22,7 @@
       #           host_path: ~/puppet
       #           container_path: /build/puppet
       #
-      if agent[:mount_folders]
+      if agent.include?(:mount_folders) && !agent[:mount_folders].empty?
```
melissa: hold on, apparently I didn't test this thoroughly enough

setup/git/010_TestSetup.rb
```diff
@@ -22,8 +22,8 @@
       #           host_path: ~/puppet
       #           container_path: /build/puppet
       #
-      if agent[:mount_folders]
-        mount = agent[:mount_folders][repository[:name]]
+      if host[:mount_folders]
+        mount = host[:mount_folders][repository[:name]]
```
melissa: it turns out, `agent` and `agents` is equivalent if you don't explicitly set `agent` to something. Who knew! The loop that this exists in we refer to the individual servers as `host`, not `agent` as I had previously, and incorrectly, assumed

melissa:
Acceptance tests have been broken for a while, and can be ignored

setup/git/010_TestSetup.rb
joshcooper: I knew this sounded familiar: https://tickets.puppetlabs.com/browse/BKR-558

PR #107, opened 2019-04-09
(maint) Added debian 10 or higher to allow insecure repo

loopinu:
(maint) Added debian 10 or higher to allow insecure repo like for ubuntu

This commit fixes the following:

When apt-get update is run against our builds our artifact repo it will throw an error saying that The repository is not signed.

By adding  the flag --allow-insecure-repository it will thrown a warning instead of an error.

More details in: https://tickets.puppetlabs.com/browse/BKR-1462

### DIFF
--- a/lib/beaker-puppet/install_utils/foss_utils.rb
+++ b/lib/beaker-puppet/install_utils/foss_utils.rb
@@ -1089,8 +1089,9 @@ module Beaker
           scp_to( host, repo, to_path )
 
           variant, version, arch, codename = host['platform'].to_array
-          if variant =~ /^ubuntu$/ && version.split('.').first.to_i >= 18
-            # Allow the use of unsigned repos with Ubuntu 18.04+
+          if (variant =~ /^ubuntu$/ && version.split('.').first.to_i >= 18) ||
+              (variant =~ /^debian$/ && version.split('.').first.to_i >= 10)
+            # Allow the use of unsigned repos with Ubuntu 18.04+ and for Debian 10 +
             on host, "echo 'Acquire::AllowInsecureRepositories \"true\";' > /etc/apt/apt.conf.d/90insecure"
           end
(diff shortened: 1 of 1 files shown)


melissa:
jenkins retest this please

melissa:
Can you remind me, why do we need to allow newer deb/ubuntu variants to use unsigned repos? It would probably be a good idea to add a bit more description along those lines in the commit message as well

loopinu:
We are doing this because newer versions of debian based os's fail if the repo is unsigned.
Before ubuntu 18.04 & debian 10 there was only a warning thrown.
Please note that this is only used to test the puppet agent and in my opinion it should be ok to use a unsecure repository.

melissa:
I agree it's probably safe in this context, but we should still have a bit more detail on why. When we look back on this in two years it would be good not to have to do a bunch of research to figure out why we made this decision. Are we not signing repos that we use for testing? Are these external repos that we're using for testing that aren't being signed? If we wanted to remove this in the future, what would be required?

You can see the previous commit that added this workaround for insecure repos (https://github.com/puppetlabs/beaker-puppet/commit/bbfc8705b7a04961ecdfb75e735073c26bf78e7f) references a beaker ticket (https://tickets.puppetlabs.com/browse/BKR-1462). Although there isn't a ton of description, we can see that it's our internal repos on build.delivery.pupetlabs.net that are unsigned. Maybe this will change with Artifactory, and we can remove this special bit of code!

Can you update the commit message with this additional detail?

melissa:
jenkins retest this please

melissa:
You'll also want to rebase your branch to pull in changes introduced with https://github.com/puppetlabs/beaker-puppet/pull/109. That'll get the PR tests passing

melissa:
I'd love to cut a new beaker-puppet release soon, and I'd love to include this change in the new release!

loopinu:
@melissa is this ok to merge now?

melissa:
I'll cut a new beaker-puppet release on Monday

PR #110, opened 2019-04-16
(maint) Restructure release repo url

melissa:


melissa:
This is in response to the email about public release repos changing. I *believe* this is the only thing that will need to change on our end. However, this can't go in until the change is made. Currently it's scheduled to go live on May 6th. The new link location does not exist yet on yum.puppetlabs.com or on nightlies.puppet.com/yum. 

I believe this change will only impact the yum repos.

We should definitely try to test this before it gets merged.

cc/ @mwaggett @jtappa @joshcooper

mwaggett:
I set up http://yum-test.puppet.com.s3-website-us-west-2.amazonaws.com/, sync'd from yum.puppetlabs.com, and will be manually restructuring it to look how we want it to after these changes. I'll send an email when that's done so you can test with that.

melissa:
For puppet tests, I realized we're never actually hitting yum.pl.com or apt.pl.com, just nightlies.pl.com or builds.d.pl.net. I have no idea about puppetserver tests, but nightlies already has the "new" layout, so this might be able to get merged in now, pending what's up with puppetserver

@Magisus @justinstoller @tvpartytonight I think this'll impact y'all as well. Care to try it out before we merge it?

melissa:
I'm so sorry this got out of control..... @caseywilliams I made some changes to your `install_puppetserver_on` method

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -1448,6 +1450,7 @@ def install_puppetserver_on(host, opts = {})
 
           if opts[:version] == 'latest' || opts[:nightlies]
             release_stream += '-nightly' unless release_stream.end_with? "-nightly"
+            opts[:version] = nil
```
jhelwig: Does this need to be `nil`, or is the point to make it "not set"? Would `opts.delete(:version)` make sense?

melissa: I think it does need to be nil. Otherwise, `install_package` below will try to pass in `latest` as the version number and it'll fail. If "not set" gets `opts[:version` to return `nil`, then that would also work.

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -1424,27 +1436,19 @@ def install_puppetserver_on(host, opts = {})
...
+          if opts[:version] == 'latest' || opts[:nightlies]
+            release_stream += '-nightly' unless release_stream.end_with? "-nightly"
+          end
+
+          # Determine the repo URLs; Use Puppet's nightly builds by default.
+          unless opts[:nightly_builds_url] == FOSS_DEFAULT_DOWNLOAD_URLS[:nightly_builds_url]
+            # If custom yum and apt urls are supplied, use them, otherwise
+            # just assume the nightly_builds_url is fine.
+            opts[:nightly_apt_repo_url] = opts[:nightly_builds_url]
+            opts[:nightly_yum_repo_url] = opts[:nightly_builds_url]
```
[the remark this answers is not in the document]
melissa: `opts[:nightly_apt_repo_url]` will always have a value. Most likely it'll be the default value. We don't have an environment variable to expose these two in the acceptance runs. It's possible the user has set them to a custom value, but unlikely from what I can tell.

It probably would be best to make this more explicit, but I wasn't sure the best way to do that and keep it simple. Do you have any thoughts?

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -374,10 +374,21 @@ def install_puppet_agent_on(hosts, opts = {})
...
+            if opts[:puppet_agent_version] && opts[:puppet_agent_version] != 'latest' && dev_builds_accessible_on?(host)
               return install_puppet_agent_from_dev_builds_on(host, opts[:puppet_agent_version])
             end
 
+            if opts[:puppet_agent_version] == 'latest'
+              opts[:puppet_collection] += '-nightly' unless opts[:puppet_collection].end_with? '-nightly'
+              opts[:puppet_agent_version] = nil
+              opts[:version] = nil
+
+              unless opts[:nightly_repo_url] == FOSS_DEFAULT_DOWNLOAD_URLS[:nightly_repo_url]
```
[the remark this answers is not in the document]
melissa: yes, but I think that's fairly unlikely. See my comment for `install_puppetserver_on`. It's the same conversation.

This is why we should have another install method that we can call from both `install_puppet_agent_on` and `install_puppetserver_on`, so we don't have to duplicate this logic in both places in slightly different ways :(

lib/beaker-puppet/install_utils/foss_utils.rb
jhelwig:
```
irb(main):001:0> a = { :version => nil }
=> {:version=>nil}
irb(main):002:0> a[:version]
=> nil
irb(main):003:0> a.delete(:version)
=> nil
irb(main):004:0> a[:version]
=> nil
```

There's no difference between the two as far as what `opts[:version]` would return, it's more about how merging the hash with other hashes, and how `opts.has_key?(:version)` would behave. There's also the slight difference in that `opts.delete(:version)` reads a bit more like "we don't care about the version whatever it was", but there aren't a ton of differences in how the code would behave.

jhelwig:
@melissa Did you want to squash the "(maint) Wow, consistency is hard" commit into where that key is introduced? Seems a little silly to introduce it with the incorrect name only to fix it later on in the series.

jhelwig:
@Magisus / @justinstoller: Any comments/thoughts on this, since it'll affect your pipelines?

justinstoller:
I've run the puppetserver smoke tests successfully with this. I'll double check using the full server and pepse test suites, next. Assuming they pass, I'll go ahead and merge.

justinstoller:
I ran the full suite and it failed due to a connection issue mid run that seemed to have nothing to do with this. Everything had installed and I did not observe any failures prior to the networking issue.

The PE tests (our suite requires this gem but I don't think uses these methods) failed, but when re-ran w/o this patch failed in the same way, so they seem unrelated (and now I have to figure out why they fail locally but pass in CI...).

PR #111, opened 2019-04-22
Maint/master/all the changes

melissa:


melissa:
tests will fail, I haven't removed associated tests yet

PR #112, opened 2019-05-03
(BKR-1587) Install non-latest builds on all hosts

joshcooper:
The non-local return caused beaker-puppet to install puppet-agent on the
first host in the `hosts` array. Use if/else to avoid the non-local
return.

### DIFF
--- a/lib/beaker-puppet/install_utils/foss_utils.rb
+++ b/lib/beaker-puppet/install_utils/foss_utils.rb
@@ -389,45 +389,45 @@ module Beaker
 
             # If inside the Puppet VPN, install from development builds.
             if opts[:puppet_agent_version] && opts[:puppet_agent_version] != 'latest' && dev_builds_accessible_on?(host, opts[:dev_builds_url])
-              return install_puppet_agent_from_dev_builds_on(host, opts[:puppet_agent_version])
-            end
-
-            if opts[:puppet_agent_version] == 'latest'
-              opts[:puppet_collection] += '-nightly' unless opts[:puppet_collection].end_with? '-nightly'
-
-              # Since we have modified the collection, we don't want to pass `latest`
-              # in to `install_package` as the version. That'll fail. Instead, if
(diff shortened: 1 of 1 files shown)


joshcooper:
To test checkout this PR, set `export BEAKER_PUPPET_VERSION=file:///Users/josh/work/beaker-puppet`, from the puppet repo, cd to `acceptance`, run `bundle update` and then:

```
$ bundle exec bundle exec rake ci:test:aio HOSTS=redhat7-64m-redhat7-64a-debian9-64a-windows2016-64a-osx1013-64a TESTS=tests/server_list_setting.rb RELEASE_STREAM=puppet6 SHA=6.4.2
```

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -389,45 +389,45 @@ def install_puppet_agent_on(hosts, opts = {})
...
-              package_name = 'puppet-agent'
-              package_name << "-#{opts[:puppet_agent_version]}" if opts[:puppet_agent_version]
-            when /debian|ubuntu|cumulus|huaweios/
-              package_name = 'puppet-agent'
-              package_name << "=#{opts[:puppet_agent_version]}-1#{host['platform'].codename}" if opts[:puppet_agent_version]
-            when /windows/
-              install_puppet_agent_from_msi_on(host, opts)
-            when /osx/
-              install_puppet_agent_from_dmg_on(host, opts)
+              install_puppet_agent_from_dev_builds_on(host, opts[:puppet_agent_version])
```
kris-bosland: It looks like you are doing an if/else block instead of the return so that we continue looping instead of exiting the method.  Can we use break instead to reduce the complexity of this block?  Also, git or something did not analyze this diff very well.

joshcooper: Mentioned this in slack, we can't use `break` because we want to continue the loop for the next host. I could use `next` instead, but meh..

PR #115, opened 2019-05-23
(maint) Mark dev repos as trusted

pcarlisle:
Instead of globally disabling package signing checks we can mark our development
repos as trusted. This allows both the repository and its packages to be
unsigned.

The previous implementation is more broad and more specific than needed, in
that it applied to all repositories and required separate settings for unsigned
repositories and unsigned packages. The setting for unsigned packages was
missing, which caused failures on the platforms that required it.

I've tested this on the oldest version of debian available in vmpooler, which
does not support this flag, and it seems to ignore the flag and still work.

### DIFF
--- a/lib/beaker-puppet/install_utils/foss_utils.rb
+++ b/lib/beaker-puppet/install_utils/foss_utils.rb
@@ -37,7 +37,7 @@ module Beaker
         # @param [Hash] the original options to be merged with the default options
         #
         # @return [Hash] The finalized set of options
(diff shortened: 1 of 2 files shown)


pcarlisle:
This should fix the currently failing puppetserver debian 8 tests.

joshcooper:
FYI @loopinu since you made a related change recently in 11744f5ae828f5b180994c67ef25ca016d097a1e

PR #116, opened 2019-06-13
(BKR-1598) Set server hostname

joshcooper:
To ensure individual tests don't need to set `--server <hostname>` set
it in the presuite for all types (git, aio, etc). Note Host#hostname is
the same as Host#to_s, but I chose the former for its explicitness.


Magisus:
This is really similar to what we do in the server presuite now 👍

PR #121, opened 2019-09-10
(maint) fix ssl cleanup

ciprianbadescu:
\* from shell command will no be expanded as it is surrounded by '

### DIFF
--- a/setup/common/040_ValidateSignCert.rb
+++ b/setup/common/040_ValidateSignCert.rb
@@ -21,7 +21,7 @@ test_name "Validate Sign Cert" do
   hosts.each do |host|
     ssldir = on(host, puppet('agent --configprint ssldir')).stdout.chomp
     # preserve permissions for master's ssldir so puppetserver can read it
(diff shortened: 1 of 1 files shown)


ciprianbadescu:
Check failures should be solved by https://github.com/puppetlabs/beaker-puppet/pull/122

setup/common/040_ValidateSignCert.rb
```diff
@@ -21,7 +21,7 @@
   hosts.each do |host|
     ssldir = on(host, puppet('agent --configprint ssldir')).stdout.chomp
     # preserve permissions for master's ssldir so puppetserver can read it
-    on(host, "rm -rf '#{ssldir}/*'")
+    on(host, "rm -rf #{ssldir}/*")
```
joshcooper: I think we added quotes originally because some older Windows versions had spaces, e.g. "Application Data". Will this still work on supported Windows platforms?

ciprianbadescu:
work continued in https://github.com/puppetlabs/beaker-puppet/pull/123 as head branch was wrong

PR #122, opened 2019-09-10
(maint) freeze activesupport < 6

ciprianbadescu:
Freeze activesupport to versions compatible with ruby 2.3 and 2.4

### DIFF
--- a/beaker-puppet.gemspec
+++ b/beaker-puppet.gemspec
@@ -30,7 +30,6 @@ Gem::Specification.new do |s|
 
   # Documentation dependencies
   s.add_development_dependency 'yard'
-  s.add_development_dependency 'markdown'
   s.add_development_dependency 'thin'
 
   # Run time dependencies


gimmyxd:
we have a dependency on `markdown` gem which depends on `textutils` which depends on `activesupport >=0`, which means it will always install the latestes activesupport version(6.0.0). Activesupport(6.0.0) is compatible only with ruby > 2.5

beaker-puppet.gemspec
```diff
@@ -31,6 +31,7 @@ Gem::Specification.new do |s|
   # Documentation dependencies
   s.add_development_dependency 'yard'
   s.add_development_dependency 'markdown'
+  s.add_development_dependency 'activesupport', '< 6.0.0'
```
joshcooper: I don't think we need the `markdown` gem at all, can we just drop it (and activesupport)?

```
$ git --no-pager grep markdown
beaker-puppet.gemspec:  s.add_development_dependency 'markdown'
```

gimmyxd: 👍 for removing if is not used

PR #125, opened 2019-11-22
(RE-12690) Add debug output to `install_repo_configs_from_url`

mwaggett:
This commit adds debug output for the repo config contents before and after
gsubbing. Filesync tests have been experiencing a transient failure where
debian repo contents are malformed. We've narrowed down the failure to this
method, somewhere between pulling down the file and scp'ing it to the host.
This debug output is intended to give us more insight into what is causing this
issue, but I haven't actually been able to reproduce the issue since making
this change. Perhaps this is enough to eliminate the transient altogether.
¯\_(ツ)_/¯

### DIFF
--- a/lib/beaker-puppet/install_utils/foss_utils.rb
+++ b/lib/beaker-puppet/install_utils/foss_utils.rb
@@ -1109,7 +1109,13 @@ module Beaker
 
           if host['platform'].variant =~ /^(ubuntu|debian)$/
             # Bypass signing checks on this repo and its packages
-            contents = File.read(repo).gsub(/^deb /, "deb [trusted=yes] ")
+            original_contents = File.read(repo)
+            puts "INFO original repo contents:"
+            puts original_contents
+            contents = original_contents.gsub(/^deb /, "deb [trusted=yes] ")
+            puts "INFO new repo contents:"
+            puts contents
+
             File.write(repo, contents)
           end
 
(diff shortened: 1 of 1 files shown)


Magisus:
Seems useful to have this debug output if the transient does happen again. I'm happy to add this.

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -1109,7 +1109,13 @@ def install_repo_configs_from_url(host, repo_config_url, copy_dir = nil)
 
           if host['platform'].variant =~ /^(ubuntu|debian)$/
             # Bypass signing checks on this repo and its packages
-            contents = File.read(repo).gsub(/^deb /, "deb [trusted=yes] ")
+            original_contents = File.read(repo)
+            puts "INFO original repo contents:"
```
Magisus: Do we have any sort of logging framework that we use in beaker, or is `puts` fine?

mwaggett: no clue lol

Magisus: Yes, we do, I see `logger.warn` elsewhere in this file. Can you switch to using that? `debug` level seems fine.

mwaggett: ah, yeah. Does the debug level show up in logs by default? I wanna make sure that we have this output if it fails in jenkins again. I'll go find that.

Magisus: Oh good point, it might not. Maybe `info` would be better.

mwaggett: The output showed up with `debug` when I ran locally, so I'm going with that.. ¯\_(ツ)_/¯

PR #127, opened 2019-12-09
(maint) Allow custom hypervisor settings

melissa:
When generating a hosts file, we should be able to pass in different
hypervisor options.


melissa:
This blocks https://github.com/puppetlabs/puppet/pull/7877

PR #132, opened 2020-02-29
Ensure that puppet runs through cmd.exe

trevor-vaughan:
Some versions of powershell do not handle streaming responses properly.
In these cases, an error will be thrown with a message like the
following: '0x5 occurred while reading the console output buffer'. Using
cmd.exe instead of powershell causes the run to work properly.


trevor-vaughan:
@genebean Referenced from office hours

trevor-vaughan:
Hmm..thought this worked, but it did not do so consistently, still trying to figure this one out.

trevor-vaughan:
This ended up being an issue in `puppet-windowsfeature` https://github.com/voxpupuli/puppet-windowsfeature/pull/136

PR #133, opened 2020-03-24
(maint) Enable Windows hosts to use the package_proxy

markcassidy:
There does not appear to be any support to enable the installation of a Windows package via the `:package_proxy` setting. This contribution would add that support.

### DIFF
--- a/lib/beaker-puppet/install_utils/foss_utils.rb
+++ b/lib/beaker-puppet/install_utils/foss_utils.rb
@@ -677,8 +677,8 @@ module Beaker
 
             if host.is_cygwin?
               # NOTE: it is critical that -o be before -O on Windows
-              proxy = opts[:package_proxy] ? " -x #{opts[:package_proxy]}" : ''
-              on host, "curl#{proxy} -o \"#{msi_download_path}\" -O #{link}"
+              proxy = opts[:package_proxy] ? "-x #{opts[:package_proxy]} " : ''
+              on host, "curl #{proxy}-o \"#{msi_download_path}\" -O #{link}"
 
               #Because the msi installer doesn't add Puppet to the environment path
               #Add both potential paths for simplicity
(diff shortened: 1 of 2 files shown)


lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -677,15 +677,20 @@ def install_a_puppet_msi_on(hosts, opts)
 
             if host.is_cygwin?
               # NOTE: it is critical that -o be before -O on Windows
-              on host, "curl -o \"#{msi_download_path}\" -O #{link}"
+              proxy = opts[:package_proxy] ? " -x #{opts[:package_proxy]}" : ''
+              on host, "curl --connect-timeout 5 --retry 20 --retry-delay 10#{proxy} -o \"#{msi_download_path}\" -O #{link}"
```
ciprianbadescu: we should add only the proxy option, without hardcoded `--connect-timeout 5 --retry 20 --retry-delay 10`

markcassidy: Sure, i'll make the adjustment now.

markcassidy: Done.

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -677,15 +677,20 @@ def install_a_puppet_msi_on(hosts, opts)
 
             if host.is_cygwin?
               # NOTE: it is critical that -o be before -O on Windows
-              on host, "curl -o \"#{msi_download_path}\" -O #{link}"
+              proxy = opts[:package_proxy] ? " -x #{opts[:package_proxy]}" : ''
+              on host, "curl#{proxy} -o \"#{msi_download_path}\" -O #{link}"
```
ciprianbadescu: I would move space to curl line for better readability
```
              proxy = opts[:package_proxy] ? "-x #{opts[:package_proxy]}" : ''
              on host, "curl #{proxy} -o \"#{msi_download_path}\" -O #{link}"

```

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -677,15 +677,20 @@ def install_a_puppet_msi_on(hosts, opts)
...
               #NOTE - this is unnecessary if the host has been correctly identified as 'foss' during set up
               puppetbin_path = "\"/cygdrive/c/Program Files (x86)/Puppet Labs/Puppet/bin\":\"/cygdrive/c/Program Files/Puppet Labs/Puppet/bin\""
               on host, %Q{ echo 'export PATH=$PATH:#{puppetbin_path}' > /etc/bash.bashrc }
             else
-              on host, powershell("$webclient = New-Object System.Net.WebClient;  $webclient.DownloadFile('#{link}','#{msi_download_path}')")
+              if opts[:package_proxy]
+                on host, powershell("$webclient = New-Object System.Net.WebClient; $webclient.Proxy = New-Object System.Net.WebProxy('#{opts[:package_proxy]}',$true); $webclient.DownloadFile('#{link}','#{msi_download_path}')")
+              else
+                on host, powershell("$webclient = New-Object System.Net.WebClient; $webclient.DownloadFile('#{link}','#{msi_download_path}')")
+              end
```
ciprianbadescu: I find the one-liner approach you used on curl better:
```
proxy = opts[:package_proxy] ? "$webclient.Proxy = New-Object System.Net.WebProxy('#{opts[:package_proxy]}',$true); " : ''
on host, powershell("$webclient = New-Object System.Net.WebClient; #{proxy} $webclient.DownloadFile('#{link}','#{msi_download_path}')")
```

lib/beaker-puppet/install_utils/foss_utils.rb
markcassidy: I understand the concern, but this will create an unnecessary space in the event the package proxy isn't set. I can shift the leading space to the end of the the `proxy` variable string as an attempt to improve readability.

lib/beaker-puppet/install_utils/foss_utils.rb
markcassidy: Done.

PR #135, opened 2020-04-30
(maint) Add extra escape chars in msi-log

donoghuc:
The powershell invocation in log_file_on uses a powershell helper that needs additional escaping for log file.


highb:
Jenkins test this please

PR #136, opened 2020-05-05
Bump puppet and rake version

highb:


highb:
Jenkins test this please

highb:
I don't think this should have an effect on anything outside of our testing pipelines so I'm going to self-merge in the interest of speed.

PR #142, opened 2020-07-27
Add .travis.yml

genebean:
This is in preparation for more Beaker stuff moving to Vox Pupuli like [core Beaker](https://github.com/voxpupuli/beaker) did today


genebean:
Agreed. I'll fix that.

PR #144, opened 2020-09-09
Add release and PR testing Github Actions

lucywyman:
This adds 2 Github Actions, copied from the Beaker repo. The first is a
release workflow that is triggered on tag to beaker-puppet, which
builds and publishes the beaker-puppet gem. The second is a per-PR
testing workflow that runs the spec tests.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,28 @@
+name: Test
+
+on: pull_request
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby:
+          - "2.4"
+          - "2.5"
+          - "2.6"
+          - "2.7"
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Ruby ${{ matrix.ruby }}
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+          ruby-version: ${{ matrix.ruby }}
+      - name: Install dependencies
+        run: bundle install
+      - name: Build docs
+        run: bundle exec rake yard
+      - name: Run tests
+        run: bundle exec rake test
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -20,8 +20,6 @@ jobs:
         with:
           bundler-cache: true
           ruby-version: ${{ matrix.ruby }}
-      - name: Install dependencies
-        run: bundle install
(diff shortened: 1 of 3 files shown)


.github/workflows/pr.yml
```diff
@@ -0,0 +1,28 @@
...
+          - "2.7"
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Ruby ${{ matrix.ruby }}
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+          ruby-version: ${{ matrix.ruby }}
+      - name: Install dependencies
+        run: bundle install
```
ekohl, line 24: Turns out this is redundant when using `bundler-cache`
```suggestion
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,25 @@
...
+          ruby-version: '2.6'
+      - name: Build gem
+        run: gem build *.gemspec
+      - name: Setup credentials
+        run: |
+          mkdir -p ~/.gem
+          echo -e "---\n:rubygems_api_key: ${{ secrets.RUBYGEMS_AUTH_TOKEN }}" > ~/.gem/credentials
+          chmod 0600 ~/.gem/credentials
+      - name: Publish gem
+        run: gem push *.gem
```
ekohl, line 25: When using Rubygems 3.0.5 or higher, you can use an env var
```suggestion
      - name: Publish gem
        run: gem push *.gem
        env:
          GEM_HOST_API_KEY: '${{ secrets.RUBYGEMS_AUTH_TOKEN }}'
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    if: github.repository == 'voxpupuli/beaker-puppet'
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Ruby 2.6
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '2.6'
```
ekohl, line 16: Ruby 2.7 contains Rubygems 3.1.4
```suggestion
      - name: Install Ruby 2.7
        uses: ruby/setup-ruby@v1
        with:
          ruby-version: '2.7'
```

ekohl (changes requested):

ekohl:
Replaced by https://github.com/voxpupuli/beaker-puppet/pull/148

PR #145, opened 2020-09-21
(maint) redhat-8-arm64 builds access check

ciprianbadescu:


genebean (approved):

ekohl (approved):
I'm ok with this short term, but long term I'd like to move away from this. Ideally we'd have some situation where the Puppet build envs would export some env var to do this rather than relying on hostnames of target machines.

PR #147, opened 2020-10-16
(maint) Make latest the latest stable instead of nightly

GabrielNagy:
Prior to this PR, if `install_puppetserver_on` was called without a version it would install the latest available nightly build.

This logic makes the use case of installing the latest *released* version unusable (i.e. from non-nightly collections), as the method
would always append `-nightly` to the collection if no version is passed.

As the `nightly` option can already be supplied to the method, change the logic to no longer assume the user wants a nightly build if they don't pass a version (or if `version: 'latest'`).

Currently, running `puppetserver_version_on` on a nightly build of puppetserver, assuming `puppetserver --version` returns something similar to:

`puppetserver version: 5.3.15.SNAPSHOT.2020.09.28T1639`

The method matches both `5.3.15` and `2020.09.28`, but returns the latter as the version, which is unusable. Return the first match instead of the last.

Did a quick check on these methods and it appears they are only used in the puppet_agent module (where I'm encountering this), so the chance of breaking unwanted stuff is minimal.

### DIFF
--- a/lib/beaker-puppet/install_utils/foss_utils.rb
+++ b/lib/beaker-puppet/install_utils/foss_utils.rb
@@ -1435,7 +1435,7 @@ module Beaker
         def install_puppetserver_on(host, opts = {})
           opts = sanitize_opts(opts)
 
-          # Default to installing latest from nightlies
+          # Default to installing latest
           opts[:version] ||= 'latest'
 
           # If inside the Puppet VPN, install from development builds.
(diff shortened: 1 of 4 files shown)


GabrielNagy:
ping @ekohl @ghoneycutt

ekohl (commented):
It looks good to me. Would a better summary be `Make latest the latest stable instead of nightly`?

GabrielNagy:
> Would a better summary be `Make latest the latest stable instead of nightly`?

Sure, definitely makes the change more descriptive. Updated!

ekohl (approved):

PR #149, opened 2020-12-18
(maint) replace use of deprecated method `get_puppet_collection`

ciprianbadescu:


ekohl (approved):
Looks like this should have been done in 4fe5b1b80aacbad9c5f50d43df686ab44380d722 which deprecated the method. IMHO your own codebase should never use deprecated methods so :+1:

PR #159, opened 2021-03-19
(maint) Update the Forge API base url

mwaggett:
The Forge API is now primarily served from the `puppet.com` domain. DNS
still resolves from the old domain, but the legacy `*.puppetlabs.com`
certificate [has been removed](https://headwayapp.co/puppet-forge-updates/removed-legacy-*-puppetlabs-com-ssl-certificate-143941).
For most users, this will have no effect because the main `puppet.com`
certificate includes `puppetlabs.com` as a SAN, but in some outdated or
otherwise broken SSL stacks, this might prevent SSL validation.

To remove that edge case, this PR updates the baseurl to the current
domain name.

### DIFF
--- a/lib/beaker-puppet/helpers/puppet_helpers.rb
+++ b/lib/beaker-puppet/helpers/puppet_helpers.rb
@@ -589,8 +589,8 @@ module Beaker
         # @param ip_spec [Hash{String=>String}] a hash containing the host to ip
         #   mappings
         # @param alias_spec [Hash{String=>Array[String]] an hash containing the host to alias(es) mappings to apply
-        # @example Stub forgeapi.puppetlabs.com on the master to 127.0.0.1 with an alias forgeapi.example.com
-        #   with_host_stubbed_on(master, {'forgeapi.puppetlabs.com' => '127.0.0.1'}, {'forgeapi.puppetlabs.com' => ['forgeapi.example.com']}) do
+        # @example Stub forgeapi.puppet.com on the master to 127.0.0.1 with an alias forgeapi.example.com
+        #   with_host_stubbed_on(master, {'forgeapi.puppet.com' => '127.0.0.1'}, {'forgeapi.puppet.com' => ['forgeapi.example.com']}) do
         #     puppet( "module install puppetlabs-stdlib" )
         #   end
         def with_host_stubbed_on(host, ip_spec, alias_spec={}, &block)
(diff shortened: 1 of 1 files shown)


lib/beaker-puppet/helpers/puppet_helpers.rb
```diff
@@ -673,12 +673,12 @@ def stub_forge_on(machine, forge_host = nil)
...
         def with_forge_stubbed_on( host, forge_host = nil, &block )
           #use global options hash
-          primary_forge_name = 'forge.puppetlabs.com'
+          primary_forge_name = 'forgeapi.puppet.com'
           forge_host ||= options[:forge_host]
           forge_ip = resolve_hostname_on(host, forge_host)
           raise "Failed to resolve forge host '#{forge_host}'" unless forge_ip
           @forge_ip ||= forge_ip
-          with_host_stubbed_on( host, {primary_forge_name => @forge_ip}, {primary_forge_name => ['forge.puppet.com','forgeapi.puppetlabs.com','forgeapi.puppet.com']}, &block )
+          with_host_stubbed_on( host, {primary_forge_name => @forge_ip}, {primary_forge_name => ['forge.puppet.com','forgeapi.puppetlabs.com','forge.puppetlabs.com']}, &block )
```
binford2k, line 681: Is this line correct?

mwaggett: @binford2k Is there something in particular that looks wrong to you? My understanding was that this should include all the aliases of the `primary_forge_name` - if I've changed the `primary_forge_name` to 'forgeapi.puppet.com', so it doesn't need to be an alias anymore, but the old value for `primary_forge_name` does. Is that not right? I'm not super familiar with beaker stuff.

Also I don't really know the difference between this method and `stub_forge_on`, but I made the same changes to both.

lib/beaker-puppet/helpers/puppet_helpers.rb
```diff
@@ -639,7 +639,7 @@ def stub_hosts(ip_spec)
         # the forge alias.
         #
         # forge api v1 canonical source is forge.puppetlabs.com
-        # forge api v3 canonical source is forgeapi.puppetlabs.com
+        # forge api v3 canonical source is forgeapi.puppet.com
```
mwaggett, line 642: I'm actually not sure if this was a change to v3 or added in a v4

ekohl (changes requested):
We've already deprecated host stubbing since it's broken and nobody should use it. How about we just remove it altogether?

https://github.com/voxpupuli/beaker-puppet/pull/160

PR #160, opened 2021-03-23
Drop forge stubbing support

ekohl:
6c87f62050c617fcf64ff7de228e2e1488ca4ee1 deprecated forge stubbing in favor of using --module-repository as a proper Puppet option. This works properly with TLS validation.

### DIFF
--- a/lib/beaker-puppet/helpers/puppet_helpers.rb
+++ b/lib/beaker-puppet/helpers/puppet_helpers.rb
@@ -589,8 +589,8 @@ module Beaker
         # @param ip_spec [Hash{String=>String}] a hash containing the host to ip
         #   mappings
         # @param alias_spec [Hash{String=>Array[String]] an hash containing the host to alias(es) mappings to apply
-        # @example Stub forgeapi.puppetlabs.com on the master to 127.0.0.1 with an alias forgeapi.example.com
-        #   with_host_stubbed_on(master, {'forgeapi.puppetlabs.com' => '127.0.0.1'}, {'forgeapi.puppetlabs.com' => ['forgeapi.example.com']}) do
+        # @example Stub host.example.com on the master to 127.0.0.1 with an alias alias.example.com
+        #   with_host_stubbed_on(master, {'host.example.com' => '127.0.0.1'}, {'host.example.com' => ['alias.example.com']}) do
         #     puppet( "module install puppetlabs-stdlib" )
         #   end
         def with_host_stubbed_on(host, ip_spec, alias_spec={}, &block)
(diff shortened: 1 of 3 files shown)


genebean:
@mwaggett - I think this might supersede the draft PR you have open

mwaggett:
thanks for the heads up @genebean! 

These methods are definitely still used in PE acceptance tests, so this has the potential to break things.

ekohl:
Given they've been deprecated for nearly 3 years now, I'd prefer to proceed with removal. It's inherently broken if you want to use TLS validation without custom CAs and issuing certificates for servers you don't own.

Can we agree on a timeframe for PE acceptance tests to migrate?

genebean:
@ekohl I think we may be able to avoid a lot of PE-related pain if you all do a major version bump on the release that includes this change. That should keep liberally configured gem files from pulling it in... does that seem right to you @mwaggett ?

ekohl:
I don't see a backwards-incompatible label, but IMHO this is certainly a major version bump since it changes the API.

genebean:
I have no doubt labels could stand to be updated .......

ekohl:
Yes, I'll see about doing that in bulk. While I haven't checked, I'm fairly certain it's true for all beaker repos that were migrated.

genebean:
I share that opinion @ekohl. Thanks!

mwaggett:
yes, this should definitely happen across a major version boundary. I _hope_ the projects that use this wouldn't automatically pull in a new major version, but I'm not really sure.

To be clear, I am in favor of this change, I'm just trying to minimize the disruption to PE teams.

ekohl:
I'm not in any hurry to merge this, but https://github.com/voxpupuli/beaker-puppet/pull/159 triggered me to submit it. I prefer not to change deprecated functionality.

Can we agree to close that PR and merge this once you give approval? That should give the PE teams sufficient time to prepare (either by pinning or migrating to the module repository option) while still moving forward.

mwaggett:
@ekohl yeah, that makes sense to me, feel free to close #159. I was hoping that PR would fix a test failure I'm seeing, but it didn't, and if the `forge` methods are potentially broken anyway, that might be why. hoping to try using the generic `host` methods soon.

since I don't have the time (nor is it my job) to audit all the PE tests, I think this is fine as-is, as long as it's released in a major version, and teams can update their pipelines if/when they break.

genebean (approved):

mwaggett:
@ekohl Is the plan to continue supporting the generic `host` equivalents of these methods (e.g. `stub_hosts_on`)?

ekohl:
Yes, they look generic and I can see use cases for that. While IMHO DNS should generally be used, I know it can be hard to guarantee in labs/dev/test/CI. That's why I didn't deprecate those APIs in my original PR.

joshcooper (changes requested):
I'm not in favor of removing these methods given the widespread usage across our CI. I think the implementation could be cleaned up, for example, we could call `Resolv.getaddress(hostname)` instead of curling and trying to match an IP address in stderr.

ekohl:
> I think the implementation could be cleaned up, for example, we could call `Resolv.getaddress(hostname)` instead of curling and trying to match an IP address in stderr.

I'm not entirely sure what you mean by that, but stubbing the forge this way is just wrong. IMHO this code shouldn't have existed in the first place.

However, recently I dove into beaker-puppet and besides a few methods (`apply_manifest_on`, `install_puppetlabs_release_repo_on`, `copy_module_to`, `install_puppet_module_via_pmt_on`, `fact_on` and `puppet_config`) it's mostly just useless to Voxpupuli's use case. I'm thinking of puling those out and getting rid of this gem in our CI. Then we can also get rid of `beaker-pe` and `beaker-puppet_install_helper` (the latter is archived but we still use it for now).

I just need to figure out the best way to avoid a conflict.

joshcooper:
> we could call `Resolv.getaddress(hostname)`

Since ruby is present on all systems under test, it's possible to run execute a ruby process to do the lookup, e.g. `ruby -e 'require "resolv"; puts Resolv.getaddress("#{hostname}")'` Or shell out to something like `nslookup`. Although it may not be installed on supported platforms.

> IMHO this code shouldn't have existed in the first place.

Yes I completely agree. Making a TLS connection to resolve an IP address and scraping stderr for the IP address is :banana:'s

> it's mostly just useless to Voxpupuli's use case. I'm thinking of puling those out and getting rid of this gem in our CI

We still use beaker-puppet in our CI systems, yet we will still need to support new puppet collections and platforms in the future. So I'd like to avoid having to fork.

binford2k:
> it's mostly just useless to Voxpupuli's use case

How can we solve this such that it meet's Vox's use case but also still Puppet's use case?

ekohl:
@joshcooper are you looking at the right issue? This is forge stubbing. The right solution to that is configuring Puppet to use a different module repository, right?

So this is getting a bit off topic, but @binford2k: there are a lot of methods that are really test cases. Like spinning up Puppetserver instance. Also there are a lot of hardcoded paths which makes sense if you can't rely on Puppet being present or you want to be explicit, but in Voxpupuli's case you generally want things to be dynamic and utilize Puppet's built in capabilities as much as possible. That means the only thing we need is to bootstrap Puppet itself. For that we need `install_puppetlabs_release_repo_on` and `install_package` (provided in Beaker itself).

Once we have Puppet present, we do utilize helpers. At least in Voxpupuli we use `copy_module_to` and `install_module_via_pmt_on` but in Foreman we've replaced those with [fixture-based installations](https://github.com/voxpupuli/voxpupuli-acceptance/blob/master/lib/voxpupuli/acceptance/fixtures.rb). `apply_manifest_on` is used a **lot** since that's how we run most tests. I think `puppet_config` (which is a simple wrapper for `puppet config print`) is used by beaker-hiera to determine the code directory.

We don't need to install from development builds nor install from git. All of that code is really complex and part of it really only works from the Puppet VPN. For our use case we don't even care about installing on Windows since we don't really test on that. Last week I started on a path where I started ripping things out and came to the conclusion that it may be easier to consider what needs to stay.

And just to be clear, I am really talking about this gem in particular (beaker-puppet).

joshcooper:
> are you looking at the right issue? This is forge stubbing. The right solution to that is configuring Puppet to use a different module repository, right?

Ah yeah, I'm somewhat conflating this with the other issue (about how to resolve the hostname into an IP). I agree the most sensible thing would be to point puppet at a different module repository, but we can't do that due that due to constraints in our infrastructure.

Also agree there is a lot of cruft in here, some of it is left over from puppet 3 and earlier days like running `puppet master` and passenger. But the forge stubbing is unfortunately still needed.

Maybe pull out the methods that you want/need into a separate gem and let this one go off into the sunset?

ekohl:
> Maybe pull out the methods that you want/need into a separate gem and let this one go off into the sunset?

That is indeed what I'm currently thinking about. Now just to figure out the naming ;)

ekohl:
We have moved away from beaker-puppet so I no longer have interest in this gem.

PR #161, opened 2021-07-15
Create better puppet apply tmpfiles

trevor-vaughan:
Adds a date-based incremental portion to the /tmp/apply_manifest*.pp files so
that debugging is easier.

Prevously, you had a set of files that were not ordered by application time.
Now, you have a set of files that is ordered by application time which means
that a simple directory listing of /tmp will tell you which one executed last.

### DIFF
--- a/lib/beaker-puppet/helpers/puppet_helpers.rb
+++ b/lib/beaker-puppet/helpers/puppet_helpers.rb
@@ -519,7 +519,7 @@ module Beaker
               puppet_apply_opts['ENV'] = opts[:environment]
             end
 
-            file_path = host.tmpfile('apply_manifest.pp')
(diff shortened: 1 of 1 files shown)


ekohl (approved):
Looks like tmpfile is implemented as `#{name}.XXXXXX` as a template, so the final filenames should sort well with this. Thanks!

On a related note: something I often found difficult is that you don't see what it's applying. Thoughts on adding a debug statement with the actual manifest content?

ekohl:
Though please have a look at the test failures :)

ekohl:
> Though please have a look at the test failures :)

Ok, so I can reproduce the failure in master so it's probably unrelated. Looks like something changed on downloads.puppet.com.

PR #164, opened 2021-08-09
Use the built-in Resolv::IPv4::Regex

ekohl:


kenyon (approved):

bastelfreak (approved):

lib/beaker-puppet/helpers/puppet_helpers.rb
```diff
@@ -12,7 +12,7 @@ module PuppetHelpers
 
         # Return the regular expression pattern for an IPv4 address
         def ipv4_regex
-          return /(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)/
+          Resolv::IPv4::Regex
```
joshcooper, line 15: Hi folks, the builtin regex uses `\A` and `\z` to match the beginning and end of the string:

```
irb(main):001:0> require 'resolv'
=> true
irb(main):002:0> Resolv::IPv4::Regex
=> /\A((?x-mi:0
               |1(?:[0-9][0-9]?)?
               |2(?:[0-4][0-9]?|5[0-5]?|[6-9])?
               |[3-9][0-9]?))\.((?x-mi:0
               |1(?:[0-9][0-9]?)?
               |2(?:[0-4][0-9]?|5[0-5]?|[6-9])?
               |[3-9][0-9]?))\.((?x-mi:0
               |1(?:[0-9][0-9]?)?
               |2(?:[0-4][0-9]?|5[0-5]?|[6-9])?
               |[3-9][0-9]?))\.((?x-mi:0
               |1(?:[0-9][0-9]?)?
               |2(?:[0-4][0-9]?|5[0-5]?|[6-9])?
               |[3-9][0-9]?))\z/
```

but the old version did not. As a result, this causes a regression in the `resolve_hostname_on` method, as we no longer match the IP address in the stderr output: https://github.com/voxpupuli/beaker-puppet/blob/3b0de16751fba4c13176193251d766c22b4a1f3b/lib/beaker-puppet/helpers/puppet_helpers.rb#L25-L28

Can you revert this change and release as 1.22.1 or 1.23.0?

PR #169, opened 2021-08-19
Prep for 1.22.1 release

tvpartytonight:


CHANGELOG.md
```diff
@@ -2,9 +2,17 @@
 
 All notable changes to this project will be documented in this file.
 
-## [v1.22.0](https://github.com/voxpupuli/beaker-puppet/tree/v1.22.0) (2021-08-17)
+## [v1.22.1](https://github.com/voxpupuli/beaker-puppet/tree/v1.22.1) (2021-08-19)
 
-[Full Changelog](https://github.com/voxpupuli/beaker-puppet/compare/1.21.0...v1.22.0)
+[Full Changelog](https://github.com/voxpupuli/beaker-puppet/compare/1.22.0...v1.22.1)
```
mwaggett, line 7: Can't put the `v` in there unless the tag will also contain a `v`, otherwise this link won't work. I don't see any other instances with `v`.
```suggestion
[Full Changelog](https://github.com/voxpupuli/beaker-puppet/compare/1.22.0...1.22.1)
```

smortex: The config says tag should start with a `v`:

https://github.com/voxpupuli/beaker-puppet/blob/0ac067e94b778bf31844a0f6f4140803261e73cd/Rakefile#L309

This was added in 0ac067e94b778bf31844a0f6f4140803261e73cd maybe we ant to adjust it?

bastelfreak: was a typo from my side. most gems we have have the prefixed `v`, but we don't want to change the existing naming scheme. Since older releases have no v we should not add it.

bastelfreak (approved):

PR #170, opened 2021-08-31
Add el8 support for install_puppet_on

gcampbell12:
beaker has support for an `el-8-*` platform but the `install_puppet_on` function does not recognise this platform

### DIFF
--- a/lib/beaker-puppet/install_utils/foss_utils.rb
+++ b/lib/beaker-puppet/install_utils/foss_utils.rb
@@ -303,7 +303,7 @@ module Beaker
             # run_in_parallel option includes install
             run_in_parallel = run_in_parallel? opts, @options, 'install'
             block_on hosts, { :run_in_parallel => run_in_parallel } do |host|
-              if host['platform'] =~ /el-(5|6|7)/
+              if host['platform'] =~ /el-(5|6|7|8)/
                 relver = $1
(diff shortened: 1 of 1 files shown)


smortex (approved):

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -303,7 +303,7 @@ def install_puppet_on(hosts, opts = options)
             # run_in_parallel option includes install
             run_in_parallel = run_in_parallel? opts, @options, 'install'
             block_on hosts, { :run_in_parallel => run_in_parallel } do |host|
-              if host['platform'] =~ /el-(5|6|7)/
+              if host['platform'] =~ /el-(5|6|7|8)/
```
ekohl, line 306: I wonder if it should assume it also works in the future (it'll likely work on EL9 too). Perhaps even ignore EL 4 and older. Nobody cares about those anymore.
```suggestion
              if host['platform'] =~ /el-(\d+)/
```

You could even merge it with fedora below if you also match on the first part.

gcampbell12: This makes sense to me, have pushed a change to merge with the fedora option as suggested

ekohl (approved):

ekohl:
@smortex want to recheck?

smortex:
:+1:

PR #173, opened 2022-02-15
(maint) Update agent download from URL logic to handle redirects

cthorn42:
Recent internal changes have caused http URLs to be redirected to HTTPs
for nightlies.puppet.com. This breaks the agent download from URL, but
this PR adds in some redirect logic that will fix the issue.

### DIFF
--- a/lib/beaker-puppet/install_utils/foss_utils.rb
+++ b/lib/beaker-puppet/install_utils/foss_utils.rb
@@ -761,9 +761,9 @@ module Beaker
                 raise "You need to specify versions for OSX host\n eg. install_puppet({:version => '3.6.2',:facter_version => '2.1.0',:hiera_version  => '1.3.4',})"
               end
 
-              on host, "curl -O #{opts[:mac_download_url]}/puppet-#{puppet_ver}.dmg"
-              on host, "curl -O #{opts[:mac_download_url]}/facter-#{facter_ver}.dmg"
-              on host, "curl -O #{opts[:mac_download_url]}/hiera-#{hiera_ver}.dmg"
+              on host, "curl --location --remote-name #{opts[:mac_download_url]}/puppet-#{puppet_ver}.dmg"
+              on host, "curl --location --remote-name #{opts[:mac_download_url]}/facter-#{facter_ver}.dmg"
+              on host, "curl --location --remote-name #{opts[:mac_download_url]}/hiera-#{hiera_ver}.dmg"
 
               host.install_package("puppet-#{puppet_ver}")
               host.install_package("facter-#{facter_ver}")
@@ -815,7 +815,7 @@ module Beaker
 
             pkg_name = "puppet-agent-#{agent_version}*"
             dmg_name = "puppet-agent-#{agent_version}-1.osx#{version}.dmg"
-            on host, "curl -O #{download_url}/#{dmg_name}"
+            on host, "curl --location --remote-name #{download_url}/#{dmg_name}"
 
             host.install_package(pkg_name)
 
@@ -826,7 +826,7 @@ module Beaker
         # Returns the latest puppet-agent version number from a given url.
         #
         # @param [String] url         URL containing list of puppet-agent packages.
-        #                             Example: https://downloads.puppetlabs.com/mac/10.11/PC1/x86_64
+        #                             Example: https://downloads.puppetlabs.com/mac/puppet7/10.15/x86_64/
         #
         # @return [String] version    puppet-agent version number (e.g. 1.4.1)
         #                             Empty string if none found.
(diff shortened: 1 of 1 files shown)


lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -834,8 +834,16 @@ def install_puppet_agent_from_dmg_on(hosts, opts)
...
           require 'net/http'
-          body = Net::HTTP.get(URI.parse("#{url}/index_by_lastModified_reverse.html"))
-          document = Oga.parse_html(body)
+
+          full_url = "#{url}/index_by_lastModified_reverse.html"
+          response = Net::HTTP.get_response(URI(full_url))
+
+          # Redirect following
+          while response.code == '301'
+            response = Net::HTTP.get_response(URI.parse(Net::HTTP.get_response(URI(full_url))['location']))
```
joshcooper, line 843: It's possible the client could get into an infinite redirect loop if the server is misbehaving. Might want to keep track of a redirect limit and then abort if we reach it, like https://stackoverflow.com/questions/6934185/ruby-net-http-following-redirects

cthorn42: Great point!

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -761,9 +761,9 @@ def install_puppet_from_dmg_on( hosts, opts )
                 raise "You need to specify versions for OSX host\n eg. install_puppet({:version => '3.6.2',:facter_version => '2.1.0',:hiera_version  => '1.3.4',})"
               end
 
-              on host, "curl -O #{opts[:mac_download_url]}/puppet-#{puppet_ver}.dmg"
-              on host, "curl -O #{opts[:mac_download_url]}/facter-#{facter_ver}.dmg"
-              on host, "curl -O #{opts[:mac_download_url]}/hiera-#{hiera_ver}.dmg"
+              on host, "curl -LO #{opts[:mac_download_url]}/puppet-#{puppet_ver}.dmg"
+              on host, "curl -LO #{opts[:mac_download_url]}/facter-#{facter_ver}.dmg"
+              on host, "curl -LO #{opts[:mac_download_url]}/hiera-#{hiera_ver}.dmg"
```
bastelfreak, line 766: in case youn want to update it: I think long options are always better. open source libs are read by a lot of people. since you already touch those lines, you could also switch to long options.

cthorn42: I updated to add the long option for `location`, but now that I think about it were you suggesting I also replace `-O` for `--remote-name` ?

smortex: > I updated to add the long option for `location`, but now that I think about it were you suggesting I also replace `-O` for `--remote-name` ?

I think it's a good idea to do so :+1:

joshcooper:
I think the ruby 3.0.3 failure is due to https://github.com/rspec/rspec-expectations/issues/1350

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -826,16 +826,28 @@ def install_puppet_agent_from_dmg_on(hosts, opts)
...
           require 'net/http'
-          body = Net::HTTP.get(URI.parse("#{url}/index_by_lastModified_reverse.html"))
-          document = Oga.parse_html(body)
+
+          full_url = "#{url}/index_by_lastModified_reverse.html"
+          response = Net::HTTP.get_response(URI(full_url))
+          counter = 0
+
+          # Redirect following
+          while response.code == '301' && counter < 15
```
joshcooper, line 843: There are a few different redirection HTTP codes, so you could check for each one, something like `[301, 302, 307, ...].include?(response.code)` or check the response class, e.g. `if response.is_a?(Net::HTTPRedirection)`

lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -826,16 +826,28 @@ def install_puppet_agent_from_dmg_on(hosts, opts)
...
           require 'net/http'
-          body = Net::HTTP.get(URI.parse("#{url}/index_by_lastModified_reverse.html"))
-          document = Oga.parse_html(body)
+
+          full_url = "#{url}/index_by_lastModified_reverse.html"
+          response = Net::HTTP.get_response(URI(full_url))
+          counter = 0
+
+          # Redirect following
+          while response.code == ('301' || '302' || '303' || '304' || '305' || '306' || '307' || '308') && counter < 15
```
joshcooper, line 843: One possible simplification, all of the redirect related responses [inherit from `Net::HTTPRedirection`](https://github.com/ruby/ruby/blob/06dab12717a05bf49e4c6dff1d6f5458678d0e00/lib/net/http/responses.rb#L73-L99) so you could check if the response is an instance of that base class:

```suggestion
          while response.is_a?(Net::HTTPRedirection) && counter < 15
```

cthorn42: Much more elegant!

joshcooper (approved):
LGTM @bastelfreak or @ekohl any concerns about this?

bastelfreak (approved):

PR #174, opened 2022-02-17
(maint) Update specs to fix ruby 3 spec failures

cthorn42:
Ruby 3 changed the way positional and keyword arguments work. In order
to get specs to pass when we pass in a hash we need to use braces. This PR
addresses that.


cthorn42:
This will fix the failures we're seeing on https://github.com/voxpupuli/beaker-puppet/pull/173.

joshcooper (approved):

PR #190, opened 2022-05-19
(PA-4178) Only clear and add non-default gem sources

joshcooper:
Rubygems 3.1 and up will prompt when adding a gem source that is too similar to
rubygems.org to protect against typo squatting. This causes tests to fail if
`GEM_SOURCE` is not overridden:

    gem source --add https://rubygems.org/
      ERROR:  While executing gem ... (Gem::OperationNotSupportedError)
        Not connected to a tty and no default specified

Since rubygems always includes https://rubygems.org/ as the default source, we
now only clear and add the source when it's not the default.

This issue was introduced due to changes in rubygems/bundler to prevent typo
squatting[1]. However, it is also triggered if the source matches `rubygems.org`
exactly.

[1] https://github.com/rubygems/rubygems/commit/aa967b85dd96bbfb350f104125f23d617e82a00a

### DIFF
--- a/lib/beaker-puppet/install_utils/foss_utils.rb
+++ b/lib/beaker-puppet/install_utils/foss_utils.rb
@@ -1625,12 +1625,18 @@ module Beaker
         # Configures gem sources on hosts to use a mirror, if specified
         # This is a duplicate of the Gemfile logic.
         def configure_gem_mirror(hosts)
-          gem_source = ENV['GEM_SOURCE'] || 'https://rubygems.org'
+          gem_source = ENV['GEM_SOURCE']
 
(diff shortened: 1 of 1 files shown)


joshcooper:
The `configure_gem_mirror` method is only called from setup steps and has no tests:

```
$ git grep configure_gem     
lib/beaker-puppet/install_utils/foss_utils.rb:        def configure_gem_mirror(hosts)
setup/common/012_Finalize_Installs.rb:  configure_gem_mirror(hosts)
setup/git/000_EnvSetup.rb:  configure_gem_mirror(agents)
```

ekohl (approved):

PR #193, opened 2022-06-23
Actually print the versions.txt file

joshcooper:
The powershell command to test if the file exists and get its contents did not
work correctly due to spaces, resulting in errors like:

    $ powershell.exe -ExecutionPolicy Bypass -InputFormat None -NoLogo -NoProfile -NonInteractive -Command Test-Path "${env:ProgramFiles}/Puppet Labs/puppet/misc/versions.txt"
      Test-Path : A positional parameter cannot be found that accepts argument 'Labs/puppet/misc/versions.txt'.

Using powershell to read the file contents is slow and requires multiple levels
of escaping so that `${env:ProgramFiles}` is not interpolated and that spaces
are handled correctly. Now it prints:

    $ cmd /c type '%PROGRAMFILES%\Puppet Labs\puppet\misc\versions.txt'
    cleanup
    facter 4.2.10-12-gb29fe9965
    hiera 3.9.0-2-g84dbdb6
    ...

### DIFF
--- a/lib/beaker-puppet/install_utils/windows_utils.rb
+++ b/lib/beaker-puppet/install_utils/windows_utils.rb
@@ -177,11 +177,13 @@ module Beaker
             # emit the misc/versions.txt file which contains component versions for
             # puppet, facter, hiera, pxp-agent, packaging and vendored Ruby
             [
-              '"${env:ProgramFiles}/Puppet Labs/puppet/misc/versions.txt"',
-              '"${env:ProgramFiles(x86)}/Puppet Labs/puppet/misc/versions.txt"'
+              "'%PROGRAMFILES%\\Puppet Labs\\puppet\\misc\\versions.txt'",
+              "'%PROGRAMFILES(X86)%\\Puppet Labs\\puppet\\misc\\versions.txt'",
             ].each do |path|
-              if file_exists_on(host, path)
-                logger.info(file_contents_on(host, path)) && break
+              result = on(host, "cmd /c type #{path}", :accept_all_exit_codes => true)
+              if result.exit_code == 0
+                logger.info(result.stdout)
+                break
               end
             end
           end
--- a/spec/beaker-puppet/install_utils/puppet_utils_spec.rb
+++ b/spec/beaker-puppet/install_utils/puppet_utils_spec.rb
@@ -153,7 +153,9 @@ describe ClassMixedWithDSLInstallUtils do
 
   describe '#puppet_collection_for' do
     it 'raises an error when given an invalid package' do
-      expect { subject.puppet_collection_for(:foo, '5.5.4') }.to raise_error
+      expect {
+        subject.puppet_collection_for(:foo, '5.5.4')
+      }.to raise_error(RuntimeError, /package must be one of puppet_agent, puppet, puppetserver/)
     end
 
     context 'when the :puppet_agent package is passed in' do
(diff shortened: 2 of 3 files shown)


lib/beaker-puppet/install_utils/windows_utils.rb
```diff
@@ -176,9 +176,17 @@ def install_msi_on(hosts, msi_path, msi_opts = {}, opts = {})
...
+            # the uninterpolated ProgramFiles env variable is passed to powershell and the
+            # result must be quoted due to spaces:
+            #
+            # 1. bash.exe -c "powershell.exe -Command Test-Path \"\\\"\${env:ProgramFiles}/Puppet Labs\\\"\""
+            # 2. powershell.exe -Command Test-Path "\"${env:ProgramFiles}/Puppet Labs\""
+            # 3. <interpreter> "${env:ProgramFiles}/Puppet Labs"
             [
-              '"${env:ProgramFiles}/Puppet Labs/puppet/misc/versions.txt"',
-              '"${env:ProgramFiles(x86)}/Puppet Labs/puppet/misc/versions.txt"'
+              '"\"\${env:ProgramFiles}/Puppet Labs/puppet/misc/versions.txt\""',
```
smortex, line 188: This syntax makes me think a lot: in a single quoted string, `\` can be used to escape a `'` or a `\` only [according to ruby-doc](https://ruby-doc.org/core-3.1.2/doc/syntax/literals_rdoc.html#label-String+Literals):
> In addition to disabling interpolation, single-quoted strings also disable all escape sequences except for the single-quote (\') and backslash (\\).

So `\"` and `\$` are "invalid" escape sequences that return the expected combination of 2 chars, but should be (I think) written `\\"` and `\\$` to avoid ambiguity (a literal `\` followed by a `"` or a `$`), in other word:

```suggestion
              '"\\"\\${env:ProgramFiles}/Puppet Labs/puppet/misc/versions.txt\\""',
```

Or maybe I am the only one confused by these unexpected quotes?

joshcooper: Yeah fun times with quotes... I'm relying on the (outer) single quoted string to disable escaping, so `\$` is the two character sequence `<backslash>, <dollar>`. Note that escaping the backslash in a single quoted string `'\\' is the same as a literal backslash (except if it's the last character). So what you're proposing is the same as what I have:

```
irb(main):004:0> '"\"\${env:ProgramFiles}/Puppet Labs/puppet/misc/versions.txt\""' == '"\\"\\${env:ProgramFiles}/Puppet Labs/puppet/misc/versions.txt\\""'
=> true
```

But I'm happy to update if you think it'd be clearer?

smortex: I *do* think it is clearer to be explicit, it is my point; but maybe I am the only one to think that way. Not sure what is the best to do with it nor if there is some best-practice regarding this :thinking: 

Both syntaxs are indeed equivalent and I think I can cope with both of them in the end.

joshcooper: I'm still not convinced that running powershell is actually necessary or desirable... I thought we used to just do `cmd /c type "%PROGRAMFILES%\......\versions.txt"` and it just worked...

joshcooper: I changed this to use cmd instead, which avoids the `$` issue and eliminates a level of escaping. Also switched to using `"` as the outer quote which requires backslashes to be escaped so it's explicit.

smortex (approved):

PR #194, opened 2022-07-08
(PA-4552) Create ci:test:setup step for iterative workflow

joshcooper:
It's now possible to setup host(s) in preparation for puppet agent related
beaker tests using:

    SHA=<SHA> HOSTS=redhat7-64a bundle exec rake ci:test:setup

SHA may be a tag or a full 40 digit git SHA.

HOSTS may be specified as a beaker-hostgenerator string or a file to an existing
hosts file. If not a file, then the abs hypervisor will be used, similar to the
`ci:test:aio` task.

For facter and puppet-agent tests, HOSTS defaults to a single redhat7-64a host.
For pxp-agent and puppet tests, HOSTS defaults to a redhat7-64m server and
redhat7-64a agent. The `SERVER_VERSION` environment variable can be used to
select a specific puppetserver version, otherwise the last released version will
be used.

When setup is complete, run 'bundle exec beaker exec path/to/test.rb' to test
iteratively.

There is a bug in beaker-abs that prevents hosts from being destroyed via
`bundle exec beaker destroy`. They must be destroyed manually for now using
floaty.

### DIFF
--- a/tasks/ci.rake
+++ b/tasks/ci.rake
@@ -25,6 +25,7 @@ Usage: bundle exec rake <target> [arguments]
 
 where <target> is one of:
 
+  ci:test:setup
   ci:test:git
   ci:test:aio
   ci:test:gem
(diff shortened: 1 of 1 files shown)


tasks/ci.rake
```diff
@@ -205,6 +205,46 @@ EOS
...
+      sha = ENV['SHA']
+      case sha
+      when /^\d+\.\d+\.\d+$/
+        # tags are ok
+      when /^[0-9a-f]{40}$/
+        # full SHAs are ok
+      else
+        puts "Error: Expected SHA to be a tag or 40 digit SHA, not '#{sha}'"
+        exit 1
+      end
```
joshcooper, line 227: This check should probably be moved to `ci:check_env` task (as long as it doesn't break existing usages for `rake ci:test:aio` or others)

joshcooper:
I really dislike using rake tasks for this purpose. The environment variables are arbitrary and auto completion doesn't work. A future improvement might be to extend `bin/beaker-puppet` with a proper CLI, e.g. `bundle exec beaker-puppet setup --hosts ... --sha  ...`

bastelfreak (approved):

PR #195, opened 2022-07-20
(PA-4566) Retry beaker exec if presuite completes

joshcooper:
Previously, if the beaker-puppet presuite or project-specific presuite
steps failed, then beaker_retry_suite would try to get a list of failed tests
and fail with:

    NoMethodError: undefined method `length' for nil:NilClass

This was because the test results file was empty, so tests_to_rerun was nil.

Now we only attempt to retry "beaker exec" if the presuite steps finished and
the list of failed tests is not empty.

If the list of failed tests is empty, then we reraise the exception. This seems
safer than blindly retrying, not knowing if beaker didn't run any tests or it
ran *all* of them and failed to write out the test results file.

### DIFF
--- a/tasks/ci.rake
+++ b/tasks/ci.rake
@@ -285,17 +285,22 @@ def beaker_suite_retry(type)
   beaker(:provision)
 
   begin
-    json_results_file = Tempfile.new
-
     beaker(:exec, 'pre-suite', '--preserve-state', '--pre-suite', pre_suites(type))
     beaker(:exec, 'pre-suite', '--preserve-state')
-    beaker(:exec, ENV['TESTS'], '--test-results-file', json_results_file.path)
-  rescue RuntimeError
-    tests_to_rerun = JSON.load(File.read(json_results_file.path))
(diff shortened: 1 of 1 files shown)


joshcooper:
```
$ cat <<END > tests/raises.rb 
raise "whoops"
END

$ SHA=7.17.0 HOSTS=redhat7-64a TESTS=tests/raises.rb bx rake 'ci:test:aio[true]'
...
Begin tests/raises.rb
RuntimeError: whoops
/home/josh/work/facter/acceptance/tests/raises.rb:1
...
ERROR: Command failed with status (1): [beaker exec tests/raises.rb --test-results...]
*** Retrying the following:
  tests/raises.rb
beaker exec tests/raises.rb
...
Begin tests/raises.rb
RuntimeError: whoops
/home/josh/work/facter/acceptance/tests/raises.rb:1
...
rake aborted!
Command failed with status (1): [beaker exec tests/raises.rb...]
```

mhashizume (approved):

PR #204, opened 2023-01-31
(maint) Call `exist?` instead of `exists?`

mhashizume:
The `exists?` method has long been deprecated and has been officially removed as of Ruby 3.2.0.


ekohl (commented):
I think some stubbing/mocking fails.

mhashizume:
Thanks for the feedback @ekohl . I've addressed the spec issues and added another commit to address a minor issue I also ran into while working on this.

ekohl (approved):

PR #205, opened 2023-02-01
Removes deprecated network device platforms

mhashizume:
This PR removes support/mention of both Arista and Cumulus, two operating systems for network devices which have been deprecated by Puppet for years.

At some point we should also remove Huawei and Cisco OSes as well.


ekohl (approved):
I wonder if we should also drop cumulus from beaker-hostgenerator if they're no longer used anyway.

PR #207, opened 2023-02-24
add support for recent FreeBSD versions; Drop FreeBSD 9/10 support

evgeni:


lib/beaker-puppet/install_utils/foss_utils.rb
```diff
@@ -719,6 +721,8 @@ def install_puppet_from_freebsd_ports_on( hosts, opts )
           block_on hosts do |host|
             if host['platform'] =~ /freebsd-9/
               host.install_package("puppet")
+            elsif host['platform'] =~ /freebsd-1[234]/
+              host.install_package("sysutils/puppet7")
```
evgeni, line 725: `sysutils/puppet` is old puppet 3, and obsoleted and doesn't exist as a pkg for recent FreeBSDs. we could just default to `sysutils/puppet7` on FreeBSD, yanking all that cruft.

bastelfreak: We already have a backwards-incompatible change merged. maybe just match on `/freebsd/` and drop the freebsd 9 support?

spec/beaker-puppet/install_utils/foss_utils_spec.rb
```diff
@@ -49,6 +49,12 @@ def logger
   let(:freebsdhost10)   { make_host( 'freebsdhost10', { :platform => 'freebsd-10-x64',
```
bastelfreak, line 49: if we drop freebsd 9, we can probably drop 10 as well since both are EoL

ekohl (approved):

PR #249, opened 2024-02-29
Re-add removed methods

mhashizume:
Two methods were mistakenly removed from both Beaker and beaker-puppet, breaking tests for modules.

puppet_agent_dev_package_info was removed in Beaker in voxpupuli/beaker@cdaedad under the mistaken assumption that it had already been moved over to beaker-puppet.

install_puppet_agent_dev_repo_on was removed in beaker-puppet in 22a38c8 because it relied on puppet_agent_dev_package_info, which was no longer available.

This commit restores those two methods (with some minor modifications for platforms that are no longer supported) and their associated tests.

### DIFF
--- a/lib/beaker-puppet/install_utils/foss_utils.rb
+++ b/lib/beaker-puppet/install_utils/foss_utils.rb
@@ -1195,6 +1195,117 @@ module Beaker
           configure_type_defaults_on(host)
         end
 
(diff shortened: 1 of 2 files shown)


mhashizume:
I think I misunderstood the test failures I was seeing and this isn't actually necessary. Closing out PR.

PR #253, opened 2024-03-18
Prep for 2.1.0 release

tvpartytonight:


lib/beaker-puppet/version.rb
```diff
@@ -1,3 +1,3 @@
 module BeakerPuppet
-  VERSION = '2.0.0'
+  VERSION = '2.1.0'
```
smortex, line 2: If it is only a bugfix release, it should be 2.0.1:
https://semver.org/#spec-item-6

tvpartytonight: Ahh, thanks for catching that @smortex ; in beaker-puppet's main branch this is listed as a bug fix, but this is actually a feature to enable Amazon Linux. I will fixup the `CHANGELOG.md` here to reflect that.

joshcooper (approved):

PR #257, opened 2024-05-30
Install puppet-agent directly on Ubuntu 24.04

joshcooper:
Ubuntu 24.04 ships a puppet-agent 8 package which uses different paths than ours. To ensure we install the correct puppet-agent package, install the specific puppet-agent artifact.

### DIFF
--- a/lib/beaker-puppet/install_utils/puppet5.rb
+++ b/lib/beaker-puppet/install_utils/puppet5.rb
@@ -217,7 +217,12 @@ module Beaker
 
           install_targets.each do |host|
             artifact_url, repoconfig_url = host_urls(host, build_details, base_url)
-            if repoconfig_url.nil?
+            if host.platform.variant == 'ubuntu' && host.platform.version.to_f >= 24.04
+              # install the specific artifact we built, not based on how its repos are configured
+              tmp_file = host.tmpfile('puppet-agent')
+              on(host, "curl -L --output #{tmp_file} #{artifact_url}")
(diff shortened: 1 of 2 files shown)


joshcooper:
I was able to install the expected 7.x puppet-agent package on Ubuntu 24.04. I think there's a case to be made to **always** install the puppet-agent artifact we built on all platforms instead of "let's rely on the package manager to figure it out"

```
$ cd ~/work/facter/acceptance
$ export BEAKER_PUPPET_VERSION="https://github.com/voxpupuli/beaker-puppet#install_repo"
$ export BEAKER_VERSION="~> 6.0" 
$ bundle update
$ env SHA=3eacd8cdb010b0152739539bd42d9a4ce4bd91e1 HOSTS=ubuntu2404-64a bundle exec rake ci:test:setup
...

cthorn42 (approved):

cthorn42:
Eventually I assume pe_repo will be updated for ubuntu 24.04, and that will cover if we are correctly pinning correctly.It isn't perfect, but it is okay if we're not testing that here.

skyamgarp:
I also executed jenkins ad-hoc pipeline with this change. 

https://jenkins-platform.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/

Are we expecting any specific reviews for this one? or this is good to merge?

PR #258, opened 2024-05-31
Release prep 4.1.0

joshcooper:
github_changelog_generator sorts based on tag date instead of semantic version,
so the changelog has been manually updated so that 2.1.0 comes between 2.0.0 and 3.0.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 All notable changes to this project will be documented in this file.
 
+## [4.1.0](https://github.com/puppetlabs/beaker-puppet/tree/4.1.0) (2024-05-31)
+
+[Full Changelog](https://github.com/puppetlabs/beaker-puppet/compare/4.0.0...4.1.0)
+
+**Implemented enhancements:**
+
+- Add support for Ubuntu 24.04 [\#257](https://github.com/puppetlabs/beaker-puppet/pull/257) ([joshcooper](https://github.com/joshcooper))
+
 ## [4.0.0](https://github.com/puppetlabs/beaker-puppet/tree/4.0.0) (2024-05-28)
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,9 +2,17 @@
...
-[Full Changelog](https://github.com/puppetlabs/beaker-puppet/compare/3.0.1...4.0.0)
+[Full Changelog](https://github.com/puppetlabs/beaker-puppet/compare/4.0.0...4.1.0)
+
+**Implemented enhancements:**
+
+- Add support for Ubuntu 24.04 [\#257](https://github.com/puppetlabs/beaker-puppet/pull/257) ([joshcooper](https://github.com/joshcooper))
+
+## [4.0.0](https://github.com/puppetlabs/beaker-puppet/tree/4.0.0) (2024-05-29)
+
+[Full Changelog](https://github.com/puppetlabs/beaker-puppet/compare/2.1.0...4.0.0)
```
mhashizume, line 15: It looks like the auto changelog generator got confused and is comparing 2.1.0 to 4.0.0, I assume it should be between 3.0.1 and 4.0.0? The generator also seems like it changed the release date from the 28th to the 29th.

mhashizume (approved):

CHANGELOG.md
joshcooper: yeah... changelog generator gets confused when there are parallel streams and it tries to make a linear history: 3.0.0 -> 2.1.0 -> 4.0.0 -> 4.1.0

Alternatively we could create per-branch changelog (so restrict to only changes in the current branch)?

```
$ gh changelog new --from-version 4.0.0 --next-version 4.1.0
```

joshcooper: I unfortunately had to manually tweak the changelog due to the issue mentioned in the description.

PR #262, opened 2025-01-16
(PA-7091) Vendor inifile in beaker-puppet

joshcooper:
The `inifile` gem has not been updated in 10+ years and does not run on Ruby 3.2 because it relies on String#taint. There is a o-inifile gem, but it has the same `taint` problem.

So vendor the MIT licensed `inifile` code into beaker-puppet and update it for Ruby 3.2. According to https://fossa.com/blog/open-source-licenses-101-mit-license/ in order to copy and/or modify the code we need to:

1. Provide the original copyright notice
2. A copy of the license itself

I've included both of those in `inifile.rb`

rubocop is failing because of an incompatibility with `rubocop-factory_bot` and `rubocop-rspec_rails`, see https://github.com/puppetlabs/dependency_checker/pull/72#issue-2343540667 for an example. That issue is fixed in https://github.com/puppetlabs/beaker-puppet/pull/263

### DIFF
--- a/beaker-puppet.gemspec
+++ b/beaker-puppet.gemspec
@@ -11,7 +11,7 @@ Gem::Specification.new do |s|
   s.description = 'For use for the Beaker acceptance testing tool'
   s.license     = 'Apache-2.0'
 
-  s.required_ruby_version = '>= 2.7', '< 3.2'
+  s.required_ruby_version = '>= 2.7', '< 3.4'
 
   s.files         = `git ls-files`.split("\n")
   s.test_files    = `git ls-files -- {test,spec,features}/*`.split("\n")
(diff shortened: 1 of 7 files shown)


beaker-puppet.gemspec
```diff
@@ -11,7 +11,7 @@ Gem::Specification.new do |s|
   s.description = 'For use for the Beaker acceptance testing tool'
   s.license     = 'Apache-2.0'
 
-  s.required_ruby_version = '>= 2.7', '< 3.2'
+  s.required_ruby_version = '>= 2.7', '< 3.3'
```
mhashizume, line 14: What are the issues with using Ruby >= 3.3? Is it just that we're not testing against those versions?

joshcooper: ah yeah good point, I didn't think to add that. Updated.

mhashizume (approved):

PR #278, opened 2026-08-09
(PE-45697) Fall back to the ssl check when PuppetDB's cleartext status port is unavailable

jonathannewman:
PE-45384 (puppet-enterprise-modules) and PE-44906 (puppetdb-private) disabled PuppetDB's cleartext HTTP status listener (port 8080) by default on fresh installs, as a security fix (SECVULN-1792) for an unauthenticated local endpoint that could accept `replace_catalog` commands. `sleep_until_puppetdb_started`'s nonssl status check unconditionally assumes that port is reachable and retries 120 times before giving up — on an affected host it now exhausts every retry and fails outright, even though the method already has a second, independent ssl-based liveness check immediately after it.

Confirmed live via a puppet-enterprise-modules frankenbuild (PR #2450, build `enterprise_puppet-enterprise-modules_init-multijob_frankenbuilder-main-main` #194, 2026-08-08): the monolithic-upgrade smoke job's "Upgrade Puppet Enterprise" step failed with `curl: (7) Failed to connect to localhost:8080` after exhausting all 120 retries, tracing back to this method (`beaker-pe`'s `upgrade_pe` → `generic_install` → `sleep_until_puppetdb_started`). This affects every consumer of `beaker-pe`/`beaker-puppet` performing a PE install or upgrade in acceptance testing, not just `pe_acceptance_tests` (tracked upstream as PE-45695, which fixed the two `pe_acceptance_tests` call sites with the same root cause).

**Fix:** catch the nonssl check's failure and fall through to the existing ssl check instead of raising — but only for the exact "exhausted retries" `RuntimeError` `retry_on` itself raises (matched by its known message template), so an unrelated `RuntimeError` still propagates rather than being misclassified as an intentionally-disabled port. Logs via `logger.warn` (not `logger.notify`, which can be silently dropped below `:notify` verbosity) so the fallback is always visible to an operator.

This rescue can't fully distinguish "port disabled by design" from a genuine PuppetDB startup failure on a host where the nonssl listener is still enabled (older PE, non-default config) — documented inline. That ambiguity is acceptable because the method doesn't return success either way without the ssl check also passing.

**Testing:**
- 3 new examples covering the fallback-succeeds, fallback-then-ssl-also-fails, and unrelated-error-propagates cases.
- Full suite: 355 examples, 0 failures. rubocop: no offenses.
- Full 3-pass self-review loop (code/error/test/comment/security/type agents): Pass 1 found and fixed an overly-broad rescue, a silently-filterable log call, and a comment overclaim; Pass 2 and Pass 3 both clean.
- **Not yet re-verified live** against a real frankenbuild — this PR isn't merged/released yet, and `pe_acceptance_tests`' `beaker-puppet` pin needs bumping once it is. Tracked as an open acceptance criterion on PE-45697.

PE-45697: https://perforce.atlassian.net/browse/PE-45697

**Not merging this PR yet** — opening for CI/review only per explicit instruction.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


lib/beaker-puppet/helpers/puppet_helpers.rb
```diff
@@ -662,9 +662,30 @@ def sleep_until_puppetdb_started(host, nonssl_port = nil, ssl_port = nil)
...
+            # fresh installs as of PE-45384/PE-44906 (SECVULN-1792), which is
+            # the expected way to hit this specific failure -- but this
+            # rescue can't distinguish that case from a genuine PuppetDB
+            # startup failure on a host where the nonssl listener is still
+            # enabled (older PE, non-default config). Either way, fall
+            # through to the ssl check below rather than failing here: it's
+            # a real (if weaker) confirmation of its own, and a true outage
+            # still fails there instead of being silently absorbed.
+            logger.warn("sleep_until_puppetdb_started: nonssl status check on port #{nonssl_port} did not succeed (#{e.message}), falling back to the ssl check")
+          end
```
joshcooper, line 696: IIUC, we'll try to curl the non-ssl port 120 times and then fall back to ssl, which will introduce ~ 2 minute delay because it sleeps for 1 second between retries. Can we try the ssl-port first always? Or try based on versions strings like we do in `if version_is_less(pe_ver, '2016.1.0')`?

jonathannewman: The try SSL first seems like a safe and future proof approach.

lib/beaker-puppet/helpers/puppet_helpers.rb
```diff
@@ -662,9 +662,30 @@ def sleep_until_puppetdb_started(host, nonssl_port = nil, ssl_port = nil)
...
+            retry_on(host, nonssl_status_command, { max_retries: 120 })
+          rescue RuntimeError => e
+            # Only treat this as the known "exhausted all retries" failure
+            # retry_on itself raises (a plain RuntimeError with this exact
+            # templated message, see beaker's Beaker::DSL::Helpers::HostHelpers
+            # #retry_on) -- not any other RuntimeError that might surface from
+            # deeper in the retry loop (e.g. an SSH/connection error from the
+            # underlying on() call). Re-raise anything else unrecognized
+            # rather than silently reclassifying it below (PE-45697).
+            raise unless e.message.match?(/\ACommand `.*` failed\.\z/m)
```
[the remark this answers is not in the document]
jonathannewman: Fixed — replaced the broad `/\ACommand `.*` failed\.\z/m/` match with an exact comparison against `nonssl_status_command`'s own templated message, so a differently-commanded templated RuntimeError from deeper in the retry loop now propagates instead of being swallowed. Added a spec case ("when retry_on raises a RuntimeError formatted like the templated failure but for a different command") to lock that in. Pushed in 8fd260c.

lib/beaker-puppet/helpers/puppet_helpers.rb
jonathannewman: Good catch — fixed by swapping the order: this now checks ssl liveness first (it's always enabled, and it's what the method's return value already reflects), then treats the nonssl check as a quick best-effort secondary confirmation with a much smaller retry budget (120 → 5), since by the time it runs, ssl has already proven the service is up. That collapses the worst case from ~180s (120 nonssl retries + ssl's own retry budget) down to roughly ssl's retry budget alone, and it works for every pe_ver without needing a version gate against an unreleased PE line. Didn't go with the version-string route (your second suggestion) because 2025.12.x hasn't shipped yet — no stable version number to gate on, and dev-build version strings don't sort predictably against a hardcoded threshold. Pushed in 8fd260c; updated specs to assert the new ssl-then-nonssl ordering. Let me know if you'd rather see something else.

lib/beaker-puppet/helpers/puppet_helpers.rb
```diff
@@ -662,11 +662,40 @@ def sleep_until_puppetdb_started(host, nonssl_port = nil, ssl_port = nil)
...
+          # The ssl status port is always enabled and is the check this method
+          # ultimately reports the result of, so use it as the primary
+          # liveness gate. Checking the nonssl port FIRST (as this method used
+          # to) meant every host with it disabled -- the default for fresh
+          # installs as of PE-45384/PE-44906 (SECVULN-1792) -- burned through
+          # 120 failed retries (~2 minutes, at retry_on's default 1s
+          # retry_interval) before ever reaching the check that actually
+          # matters (PE-45697).
+          result = curl_with_retries('start puppetdb (ssl)',
+                                     host, "https://#{host.node_name}:#{ssl_port}", [35, 60])
```
joshcooper, line 674: what's the significance of 35 and 60? Maybe use `curl -k` so we can avoid checking for specific SSL errors?

jonathannewman:
Re: Copilot's suppressed comment on the fallback spec (line 1066) — good catch. Confirmed by temporarily reverting `max_retries` back to 120 and re-running: the spec still passed because the block ignored `_opts`, so a regression back to the original 120-retry budget (the delay this PR removes) wouldn't have been caught. Fixed by adding `.with(anything, anything, { max_retries: 5 })` to the expectation alongside the block; re-verified it now fails against the reverted code and passes against the real fix. Pushed in 7c179bc.

lib/beaker-puppet/helpers/puppet_helpers.rb
jonathannewman: Per `man curl`: 35 = SSL connect error (handshake failed), 60 = peer certificate can't be authenticated with known CA certificates. This `[35, 60]` list is pre-existing, not something this PR introduced — I only moved this line earlier in the method, the exit codes are untouched. It's used identically in 3 other spots in this same file (`bounce_service`, `sleep_until_puppetserver_started`, `sleep_until_nc_started`), so it looks like an established beaker-puppet convention: since beaker's curl client typically doesn't have the PE root CA in its trust store, a clean exit 0 against a real mTLS endpoint may never happen, but reaching a TLS-handshake-stage error (35/60) is itself proof the service is up and terminating a handshake — which is apparently treated as sufficient liveness confirmation here.

`curl -k` would change that: skip cert verification and treat a real exit 0 as success instead of accepting 35/60. That seems like a plausible simplification, but it'd be a behavioral change to a security-relevant TLS check used in 4 places across this file, none of which are what this PR set out to fix (the nonssl/ssl check ordering). I'd rather not fold it into this PR without dedicated testing across all 4 call sites — want me to file a follow-up ticket for it, or would you rather I scope it into this PR now?

